PR #2, opened 2021-04-21
Disclaimer about compiler standards

RenHirayama:
ROOT (&ge 6.12.07?) has a `string_view` type which is only present in the C++17 standard.


akschaefer (approved):
Looks good 👍

elfnerhannah:
Can this be merged?

PR #25, opened 2023-02-07
Add targets for LHC energies

NGoetz:
I have added targets for both 2.76 and 5.02 TeV. The hydro parameters are tuned in such a way to reproduce approximately the midrapidity yield, mean midrapidity transverse momentum and v2  at 0-5% centrality, taken from 1303.0737, 1910.07678 and https://doi.org/10.1103/PhysRevLett.116.132302 respectivly.


zpauliny (approved):
Very good to have these ready!

zpauliny:
Before merging, add the new collision systems to the README, please.

PR #27, opened 2023-11-20
Fix reading of number of particles from config and commented lines.

JanHammelmann:
This PR fixes the readout of the number of particles from the config and changes the way a commented line is read to extract the spectators from the initial stage.


NGoetz (approved):
Looks good! Thank you for fixing this!

PR #45, opened 2024-11-11
FIST sampling

NGoetz:
This PR is not final yet, as notably 

- [x] additional unit tests are required
- [x] documentation of the new feature is required

I did some test-driven development for this one, so I started with the functional tests and implemented all necessary features to offer a second module for the sampler. My assumption was that it is rather unlikely that anytime soon, a third sampler module will be added. Therefore, I decided to minimize code duplication and reuse as much of the existing infrastructure as possible.
Notably, the difference between the two samplers is limited:

- Both have a two column txt config. Only the names of the fields differ
- The FIST sampler needs external particle and decay lists. I treat them like the freezeout, and create symlinks in the output folder.

I created the PR on purpose before everything is done, to already get some feedback in case some of the design decisions are suboptimal.


NGoetz:
I have extracted the functionalities for the separate samplers in separate files. I will add the documentation once a preliminary review finds that the implementation is okay. 
Regarding unit tests, do you think it is necessary to duplicate all unit tests for both samplers?

bash/global_variables.bash
```diff
@@ -154,10 +160,17 @@ function Define_Further_Global_Variables()
         [Sampler]=''
         [Afterburner]=''
     )
+    declare -gA HYBRID_module=(
+        [Sampler]='SMASH'
```
AxelKrypton, line 164: Shouldn't this be `=${HYBRID_default_sampler_module}`? Apart from here, where is the variable `HYBRID_default_sampler_module` used?

bash/sanity_checks.bash
```diff
@@ -23,6 +23,10 @@ function Perform_Sanity_Checks_On_Provided_Input_And_Define_Auxiliary_Global_Var
             __static__Ensure_Executable_Exists "${key}"
             __static__Set_Software_Configuration_File "${key}"
             __static__Set_Software_Input_Data_File_If_Not_Set_By_User "${key}"
+            if [ "${key}" = "Sampler" ]; then
```
AxelKrypton, line 26: Use bash test keyword `[[`.

bash/sanity_checks.bash
```diff
@@ -55,6 +59,33 @@ function Perform_Internal_Sanity_Checks()
 
 #===================================================================================================
 
+function __static__Ensure_Valid_Module_Given()
+{
+    if [ "${HYBRID_module[Sampler]}" != "SMASH" ] \
+        && [ "${HYBRID_module[Sampler]}" != "FIST" ]; then
```
AxelKrypton, line 65: I usually prefer to implement this as
```bash
if [[ ! "${HYBRID_module[Sampler]}" =~ ^(SMASH|FIST)$ ]]; then
```
as it is easier to generalise.

bash/sanity_checks.bash
```diff
@@ -55,6 +59,33 @@ function Perform_Internal_Sanity_Checks()
...
+        exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
+            'The module specified for the Sampler run is not valid.' \
+            'Valid modules are: ' --emph 'SMASH' ' and ' --emph 'FIST' '.'
+    fi
+    return 0
+}
+
+function __static__Ensure_Additional_Paths_Given_For_Sampler()
```
AxelKrypton, line 73: In this function I would use the `Ensure_Given_Files_Exist` utility function, to which you can pass text to be printed. E.g.
```bash
Ensure_Given_Files_Exist \
    'Some needed file for Thermal-FIST sampler was not specified in the configuration file' \
    -- "${HYBRID_fist_module[Particle_file]}" "${HYBRID_fist_module[Decays_file]}"
```

bash/source_codebase_files.bash
AxelKrypton, line 1: I would probably call the new files as `Sampler_functionality_{FIST|SMASH}.bash` as it makes developer life easier in an IDE (they appear next to each other).

bash/utility_functions.bash
```diff
@@ -589,4 +589,17 @@ function Make_Functions_Defined_In_This_File_Readonly()
     fi
 }
 
+: << 'DOCSTRING'
+--8<-- [start:Copy_Associative_Array-desc]
+Copy an associative array
+DOCSTRING
+function Copy_Associative_Array()
```
AxelKrypton, line 596: Where is this used? I probably missed it. I am asking because if we decide to offer this function we need to
* document it better, in particular the interface wanting to receive two names (a calling example should be added);
* ensure the passed names refer to associative arrays
* declare `key` as local variable and spell `source` and `destination` entirely.

tests/unit_tests_Sampler_functionality.bash
AxelKrypton, line 1: Same comment as above here. Use suffixes instead of prefixes.

bash/FIST_sampler_functionality.bash
AxelKrypton, line 1: Functions in this file should not be prefixed with `__static__` as they are used in different files. Same applies to the SMASH sampler functionality. I forgot this when giving you hints in Slack, sorry.

AxelKrypton (changes requested):
@NGoetz You did an excellent job so far. I am amazed of how few code changes are needed (very good example of the OCP).

I left around lots of comments about the implementation, I did not yet review in detail the black-box and the tests.

> Regarding unit tests, do you think it is necessary to duplicate all unit tests for both samplers?

As rule of thumb, the units to be tested are the specific functions. Then tests for the generic sampler functionality are more integration tests and I would say we can be less meticulous there. So the rationale is: "Let's test in a solid ways ingredients and have at least one meaningful test for their integration".

---

As you said documentation is missing. I guess this will require some non trivial effort and I really encourage you to dedicate enough time and energy to it. After all that is the contact point with the public and we want to keep quality there as high as possible. We might also do that in a separate PR. Actually I'd probably do so. I'd also deploy `develop` branch documentation once that PR is merged. 😏

bash/FIST_sampler_functionality.bash
```diff
@@ -0,0 +1,192 @@
...
+            'particle_list_file' "${particle_list_file}" \
+            'decays_list_file' "${decays_list_file}")"
+}
+
+function __static__Get_Surface_Path_Field_From_Sampler_Config_As_Global_Path_FIST()
+{
+    __static__Get_Path_Field_From_Sampler_Config_As_Global_Path_FIST 'hypersurface_file'
+}
+
+function __static__Get_Path_Field_From_Sampler_Config_As_Global_Path_FIST()
```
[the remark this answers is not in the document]
NGoetz: That is a great suggestion, I really like this solution!

bash/Sampler_functionality.bash
```diff
@@ -16,8 +16,10 @@ function Prepare_Software_Input_File_Sampler()
     Replace_Keys_In_Configuration_File_If_Needed_For 'Sampler'
     __static__Validate_Sampler_Config_File
     __static__Check_If_Sampler_Configuration_Is_Consistent_With_Hydro
-    __static__Transform_Relative_Paths_In_Sampler_Config_File
+    __static__Transform_Relative_Paths_In_Sampler_Config_File_For_${HYBRID_module[Sampler]}
     __static__Create_Superfluous_Symbolic_Link_To_Freezeout_File_Ensuring_Its_Existence
+    __static__Create_Superfluous_Symbolic_Link_To_External_Files_Ensuring_Their_Existence_${HYBRID_module[Sampler]}
+
```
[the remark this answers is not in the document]
NGoetz: I decided to not merge the functions in order to separate general and module dependent functionality.  As long as the majority of code is module-independent, I think this helps. Should we ever have less common code, one could consider merging them. I deleted the stray line.

bash/global_variables.bash
NGoetz: True, this was leftover from refactoring.

bash/utility_functions.bash
NGoetz: This function was used at some point during the changes but has been dropped again, thanks for catching this.

NGoetz:
I have included above comments. I will now move on to implement some unit tests. As discussed, the documentation will be part of a separate PR.

bash/sanity_checks.bash
```diff
@@ -48,13 +54,58 @@ function Perform_Internal_Sanity_Checks()
...
+{
+    if [[ "${HYBRID_module[Sampler]}" = 'FIST' ]]; then
+        Ensure_Given_Files_Exist \
+            'Some needed file for Thermal-FIST sampler was not specified in the configuration file' \
```
AxelKrypton, line 82: You (and I) missed a period at the end of the message.

bash/sanity_checks.bash
```diff
@@ -48,13 +54,58 @@ function Perform_Internal_Sanity_Checks()
...
+            -- "${HYBRID_fist_module[Particle_file]}" "${HYBRID_fist_module[Decays_file]}"
+    fi
+}
+
+function __static__Set_Sampler_Input_Key_Paths()
+{
+    declare -rgA HYBRID_sampler_input_key_default_paths=(
```
AxelKrypton, line 89: Since reading this I asked myself "Why can't this array be declared in the global variables file?", I would add here an amplification comment to explain this. If I am correct, it is because the `particle_list_file` and the `decays_list_file` could be changed by the user in the handler config file through the `HYBRID_fist_module` array. Hence, for convenience, we put here everything. Is it so or am I missing something else? 🤔

bash/sanity_checks.bash
```diff
@@ -48,13 +54,58 @@ function Perform_Internal_Sanity_Checks()
...
+        ['surface']="${HYBRID_software_output_directory[Hydro]}/freezeout.dat"
+        ['spectra_dir']="${HYBRID_software_output_directory[Sampler]}"
+    )
+}
+
+function __static__Choose_Base_Configuration_File_For_Sampler()
+{
+    if [[ "${HYBRID_software_base_config_file[Sampler]}" = '' ]]; then
+        Sampler_key="Sampler_${HYBRID_module[Sampler]}"
+        HYBRID_software_base_config_file[Sampler]="${HYBRID_software_base_config_file[${Sampler_key}]}"
```
AxelKrypton, line 105: I guess the variable is for readability. Still, I would comply with names style rules (only lowercase) and make it `local -r`.

python/requirements.txt
```diff
@@ -8,3 +8,4 @@ packaging>=17.0
 numpy~=1.11
 PyYAML~=6.0.1
 pyDOE~=0.3.8
+DateTime~=5.0
```
AxelKrypton, line 11: I have nothing against this, but I do not see where it is used. Can you tell it to me, please? 😅

tests/mocks/fist_sampler_black-box.py
AxelKrypton, line 1: I would rename the other sampler black-box script consistently to *smash_sampler_black-box.py*.

I also saw that you basically duplicated the SMASH one and changed it to make it FIST-like. I know sometimes we make so because of lack of time, but I believe one should avoid that as much as possible. I'd have preferred to have e.g. a positional command line argument to be `FIST` or `SMASH` to tune behaviour. Please, as bare minimum, open an issue to do this in the future.

tests/functional_tests_Sampler_only.bash
```diff
@@ -31,6 +81,51 @@ function Functional_Test__do-Sampler-only()
     fi
     Check_If_Software_Produced_Expected_Output 'Sampler' "$(pwd)/Sampler"
     mv 'Sampler' 'Sampler-success'
+    # Expect success with FIST module
+    mkdir -p "Sampler/${run_id}"
+    touch "${particle_list}"
+    touch "${decays_list}"
```
AxelKrypton, line 87: Touch the two files in single `touch`. Same applies later at many points in the same file.

tests/functional_tests_Sampler_only.bash
AxelKrypton, line 1: Test code should adhere the same clean code practices of normal code. For some reason (probably the way we learnt programming), people (me included) tend to be sloppier in test code. 😅 In this file each functional test is basically duplicating code. Lines 84 to 95, 96 to 106, 107 to 117, 118 to 128 are basically a call to a "static" function to be extracted. The first block has one more line since it tests for success, but you can still include it in the function. You need to pass arguments to print messages, call the handler and to test for success, but it is feasible. If you _really_ do not have time, leave it as it is and open an issue.

tests/unit_tests_Sampler_functionality.bash
```diff
@@ -69,6 +71,38 @@ function Clean_Tests_Environment_For_Following_Test__Sampler-create-input-file()
     rm -r "${HYBRID_output_directory}"
 }
 
+function Make_Test_Preliminary_Operations__Sampler-create-input-file-with-FIST()
+{
+    __static__Do_Preliminary_Sampler_Setup_Operations
+    HYBRID_module[Sampler]='FIST'
+    touch "${HYBRID_fist_module[Particle_file]}"
+    touch "${HYBRID_fist_module[Decays_file]}"
```
AxelKrypton, line 79: One `touch` only. Same elsewhere in this file.

tests/unit_tests_Sampler_functionality.bash
```diff
@@ -69,6 +71,38 @@ function Clean_Tests_Environment_For_Following_Test__Sampler-create-input-file()
...
+        return 1
+    elif [[ ! -f "${HYBRID_software_configuration_file[Sampler]}" ]]; then
+        Print_Error 'The output directory and/or software input file were not properly created.'
+        return 1
+    fi
+}
+
+function Clean_Tests_Environment_For_Following_Test__Sampler-create-input-file-with-FIST()
+{
+    HYBRID_module[Sampler]='SMASH'
```
AxelKrypton, line 100: This is suspicious here, as tests are run in sub-shells and we never deal with variables in clean-up. 🤔 I would simply set this variable in the beginning of unit tests as you do in the preliminary operations either to `SMASH` or to `FIST`.

tests/unit_tests_Sampler_functionality.bash
```diff
@@ -212,6 +246,108 @@ function Clean_Tests_Environment_For_Following_Test__Sampler-validate-config-fil
...
+        Print_Error 'Config validation passed although config file does not contain "hypersurface" key.'
+        return 1
+    fi
+    # Config file missing required key 'output_file'
+    printf '%s\n' "hypersurface $(which ls)" > "${HYBRID_software_configuration_file[Sampler]}"
+    Call_Codebase_Function_In_Subshell __static__Is_Sampler_Config_Valid &> /dev/null
+    if [[ $? -eq 0 ]]; then
+        Print_Error 'Config validation passed although config file does not contain "output_file" key.'
+        return 1
+    fi
```
AxelKrypton, line 302: There are more than two required keys, isn't it? Here the tests are just a mere duplication of SMASH sampler, while for FIST something more is needed. I think that you should put 3 out of 4 mandatory keys and repeat for the 4 possible missing keys.

tests/utility_functions_functional.bash
```diff
@@ -18,6 +18,9 @@ function Check_If_Software_Produced_Expected_Output()
         Sampler)
```
AxelKrypton, line 18: I'd rename this to `Sampler_SMASH` for consistency.

AxelKrypton (changes requested):
I finally found time to go through this PR. Although I left many comments, they are almost all minor. Those who are not are mainly about amplification comments or the DRY principle. When implementing changes like you did in this PR one is very much tempted to duplicate and adapt, which is fine, provided that one refactors out duplications once done. 😉 Just do the proposed small adjustments, I am confident we are ready to merge then.

bash/Sampler_functionality_FIST.bash
```diff
@@ -0,0 +1,163 @@
...
+    Ensure_Input_File_Exists_And_Alert_If_Unfinished "${particle_list_file}"
+    Ensure_Input_File_Exists_And_Alert_If_Unfinished "${decays_list_file}"
+    if [[ "$(dirname "${particle_list_file}")" != "${HYBRID_software_output_directory[Sampler]}" ]]; then
+        ln -s "${particle_list_file}" \
+            "${HYBRID_software_output_directory[Sampler]}/$(basename "${particle_list_file}")"
+    fi
+    if [[ "$(dirname "${decays_list_file}")" != "${HYBRID_software_output_directory[Sampler]}" ]]; then
+        ln -s "${decays_list_file}" \
+            "${HYBRID_software_output_directory[Sampler]}/$(basename "${decays_list_file}")"
+    fi
```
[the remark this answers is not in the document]
NGoetz: Much nicer, thank you for the suggestion!

bash/sanity_checks.bash
NGoetz: This is indeed the correct reason, I will point this out accordingly.

tests/mocks/fist_sampler_black-box.py
NGoetz: I think I could take care of merging the black boxes in this PR already, I think it is not too much work. I am not sure of the strategy though. A positional command line argument would need to be provided when the handler calls this executable, which is done by the `Run_Sampler_Software_` function, so we cannot enforce the positional argument for tests only. 
But it could be that I am overlooking a simple solution here

AxelKrypton: Right, I forgot we do not manually run it in the tests, but through the handler. Then you could take the information about the type of sampler in the black-box from a dedicated environment variable. In the functional tests you should set it accordingly. Basically the same strategy as for the failures.

python/requirements.txt
NGoetz: I do not remember why I put this requirement. I think I had some problem on my office machine. In any way, this problem does not occur on my laptop, and "DateTime" (in contrast to "datetime") is not used, so I removed this again.

AxelKrypton (approved):
Well done! 👍 I still have some minor comments, but the implementation looks good and you did a very good work in the mocking black-box. I like it much more now than before. 😊 I already approve, so that you can merge once the left-over adjustments are dealt with.

PR #46, opened 2024-11-19
Update to ITP email addresses

RobinSattler:
I noticed that the handler still uses the fias emails in the AUTHOR.md file and updated this. Additionally, I moved Jan Hammelmann to "Past developers". @AxelKrypton, could you double check the updated mail addresses?

I also updated the .mailmap file to the itp email addresses and added a few addresses that were missing in it.


AxelKrypton (approved):
Thanks for taking care of this! 👍

PR #47, opened 2024-11-20
Documentation for FIST sampling

NGoetz:
This is an add-on to #45, containing the documentation.  A general section on adding new modules should give an entry point to future developers.

### DIFF
--- a/docs/developer/new_module.md
+++ b/docs/developer/new_module.md
@@ -1,0 +1,12 @@
+# Adding a new module
+
+Although the SMASH-vHLLE-hybrid, as implied in the name, was built with a defined set of software modules, it is designed to be easily extensible with new modules. This allows, amongst other things, to test new algorithms and study model dependency.
+
+Adding a new module to the SMASH-vHLLE-hybrid is a straightforward process. The following steps are required:
+
+* In case the stage has no other modules, add a field to the `HYBRID_module` array in the `global_variables.bash` script. At the same place, add alternative base config files for the stages named `${STAGE_NAME}_${MODULE_NAME}` and add `[MODULE]` to the valid keys of the stage
+* Add a base configuration in the `configs` folder for the new module
+* Adapt the check for valid modules in the `sanity_checks.bash` script, and add a function to choose the base configuration file dependent on the module.
+* Add versions of the affected stage functionality script for the new module. Shared logic stays in the common functionality script, whereas module-specific logic is implemented in the module-specific script. The module dependent logic is implemented in functions with the suffix `_${MODULE_NAME}`. Therefore, if-clauses in the main stage script can be avoided.
+* Don't forget to source the new files!
+* Add a mock for the new module in the mocks folder.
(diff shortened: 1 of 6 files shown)


docs/developer/new_module.md
AxelKrypton, line 1: I like the intent of this file! 💪 

I'd refine it in the following way:

1. Use periods at the end of each bullet point. Use one sentence per line, too.
2. I would add e.g. an `info` or `tip` admonition box after the first paragraph to tell the user that the FIST implementation can be looked up to see how to add a new module in the future.
3. `:material-file: _Files_` and `:file_folder: **folders**` in the documentation are written prepending emojis and using italic/bold font. Do it here, too.
4. I would milden the sentence "The following steps are required:" and simply say that the following steps serve as guidance and are probably needed when adding a new module.
5. Maybe it makes it easier to give instructions referring to a concrete module name. You can e.g. say something like: "For the sake of clarity, let's suppose to be interested in adding the `Octopus` software as new afterburner." 😅 OK, you can use `MUSIC` for hydro maybe...
6. I would more "pedagogical" in the steps. I am not sure a reader that did not deal with the code as I did in the PR would understand what you write here. 🤔 Maybe adding a concrete example for each statement would help already, but couples code and documentation too much, IMO. What if we just describe the steps more in an abstract way, like you did for the last points?
7. `[MODULE]` should be `[Module]` I guess.
8. Although not needed, consider adding  a direct link to this page in the developer guide overview page.

AxelKrypton (changes requested):
I have suggested some polishing work and some refinements. I am happy to discuss this further, if needed. 😊

docs/user/prerequisites.md
```diff
@@ -38,6 +38,17 @@ Instructions on how to compile or install the software above can be found at the
     We particularly highlight that the newer versions of ROOT require C++17 bindings or higher, which calls for proper treatment of compiler options in SMASH and the hadron sampler.
     It is also recommended to start from a clean build directory whenever changing the compiler or linking to external libraries that were compiled with different compiler flags.
 
+In principle, the Hybrid handler is agnostic to the physics model used in each state, and is built in a way to support different software with minimal efforts. So far, the following choices exist for different stages:
```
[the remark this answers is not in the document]
NGoetz: This sounds to me like we can by construction only offer flexibility for some of the stages. But we actually can replace a lot of things, we just haven't implemented it all. How about:

`In principle, the Hybrid handler is agnostic to the physics model used in each state, and is built in a way to support different software with minimal efforts. Currently, this is realized by supporting for the sampling stage:`

docs/developer/new_module.md
NGoetz: I am still torn between making it more specific, and the wish to not couple code and documentation.
The way I went now was to add a few small snippets of code. My reasoning was the following: they are small and unlikely to change much anytime, but the next time a module is added, I will probably be not around anymore, so it makes sense to make it as easy accessible as possible. But of course we can also remove them again if you think this is not worth it.

AxelKrypton (approved):
Excellent work @NGoetz. 👏 I took the freedom to refine around the documentation, I hope you are fine with my changes. If not, speak up. If you agree with them, just go ahead and merge. I will then deploy `develop` documentation.

Ah, I forgot, you need to rebase on `develop` before merging. I do not expect conflicts, though.

PR #50, opened 2024-12-16
Fix failing test and potential pip problem

### LINKED ISSUE #49: Troubles running the Hybrid-Handler due to dependencies
I tried to run the Hybrid-Handler locally on my macOS Sequoia 15.1.1. machine and ran into several hurdles:

- As already known the handler is not compatible with the newest version of numpy. On my machine, I am using numpy 1.23.5 which is too high already, so I need to set up a virtual environment.
- When trying to create a virtual environment to pip install numpy 1.11 I realized that numpy 1.11 is not compatible with the newer python versions, which is python 3.9.7 in my case. This means that I have to create a pyenv to create a venv to run the Hybrid-Handler.

I think it would be handy to allow for higher numpy versions, at least within numpy 1.xx such that at least the last step of creating a pyenv would become obsolete. Is there a certain reason why we limit the handler to ~1.11?

With that limitation I have to say that for me personally, the usage of the handler feels quite unhandy and not super intuitive.

AxelKrypton:
This closes #49.

The fix was simple, but I did not invest time to find a way to have a test covering this. I am actually not sure we can really do it and I am also not sure we should do so, as it is something on `pip` side.

On the way I fixed another failing test, which was a naive mistake in being too strict fixing the reference value.

### DIFF
--- a/tests/unit_tests_version.bash
+++ b/tests/unit_tests_version.bash
@@ -14,7 +14,13 @@ function Make_Test_Preliminary_Operations__version()
 
 function Unit_Test__version()
 {
-    HYBRID_codebase_version='SMASH-vHLLE-hybrid-[0-9]+([.][0-9]+)?'
+    # To make sure this test is always sound, we set here a fake fall-back
+    # version number in the following variable. If Git history is available this
+    # is not used, but for example on GitHub in the self-hosted actions it will
+    # be used as the repository is locally checked out effectively as a
+    # black-box archive. Testing using a regex should cover all cases.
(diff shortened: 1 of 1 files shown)


AxelKrypton:
I am not sure what makes the actions fail, since they both pass on my machine and on the ITP desktop. 🤔 I will push a commit with debug output to be later cleaned up.

NGoetz (approved):
Looks good, thank you for implementing this fix!

nilssass:
Thank you for the fix. This solves the issue.

nilssass (approved):

PR #51, opened 2024-12-23
Modify IC base configuration file and predef configs

zpauliny:
This PR has two small changes: 
- new syntax for initial conditions in SMASH IC config is needed (for SMASH 3.2)
- turn on Monash tune in Afterburner stage at highest energies for consistency (needed to add this key to afterburner base config)


NGoetz (approved):
Tested, works!

PR #53, opened 2025-01-20
Add more instructive sentence to failure output

### LINKED ISSUE #41: Very difficult usage of the handler in Virgo
I'm trying to use it in the cluster with @EggerJoscha but currently it's very hard.

The problems we found:
- we must use a smash container to open a virtual environment to install pip to download requirements (some that are not in *python/requirements.txt*), instead of just having a container with the necessary requirements
- numpy 2.1.1 is not accepted even though it requires 1.11
- several unit tests fail, see image below (functional tests work)
- the error message is not helpful (e.g. "Fatal: IC run failed"). We found the logs after solving the issue, but it'd be helpful if the message warned us that there was a log
- "column" is not available in the container but this didn't interfere in anything so far

[tests_runner.log](https://github.com/user-attachments/files/17031867/tests_runner.log)
image

It seems that the original problem is a Pythia installation, but we took a longer time to realize this and I think the manager could've helped us. We got it working with a workaround from @RobinSattler, but at least we should have info on how to do it in our internal GitLab wiki. 

I think we need to improve this side though, as the hybrid is mostly done in HPC

AxelKrypton:
This closes #41 IMO. It is a minor improvement, but the only one I believe that we should undertake.

Of course, we should decide whether we want to ship the container that we built for internal use and, if so, we should then mention in the documentation. I will open an issue to discuss this further.

### DIFF
--- a/bash/common_functionality.bash
+++ b/bash/common_functionality.bash
@@ -1,6 +1,6 @@
 #===================================================
 #
-#    Copyright (c) 2024
+#    Copyright (c) 2024-2025
 #      SMASH Hybrid Team
 #
 #    GNU General Public License (GPLv3 or later)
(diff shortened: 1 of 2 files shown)


RenHirayama (approved):
There isn't much else to be done from the handler perspective so I approve, but I think we should make this container publicly available

Issue #52, opened 2025-01-16
Adjust smash-hadron-sampler execution because of breaking changes in sampler repository


RobinSattler:
In PR [#31](https://github.com/smash-transport/smash-hadron-sampler/pull/31) of the smash-hadron-sampler, the config key `spectra_dir`, used to specify the output directoy, was renamed to `output_dir`, and in PR [#38](https://github.com/smash-transport/smash-hadron-sampler/pull/38) the config key `surface`, used to specify the freezeout file, was renamed to `surface_file`. Additionally, the command line parameters have been changed from `events` and `NUM` to

    -c, --config <file>         Mandatory parameter to specify the config file.
    -n, --num <integer>         Optional number to create a random seed, useful to run several
                                instances of the sampler in parallel. The number is also the
                                ROOT output filename.
    -o, --output <directory>    Optional parameter to overwrite the output directory given
                                by "output_dir" in the config file.
    -s, --surface <file>        Optional parameter to overwrite the hypersurface freezeout
                                file given by "surface" in the config file.

Additionally, there were some config keys in the sampler that were initialized but never used. These were removed in [PR #28](https://github.com/smash-transport/smash-hadron-sampler/pull/28). And I noticed that the config key `create_root_output` is not a valid key in the hybrid handler. Furthermore, I think in the case of the smash-hadron-sampler we should ship a config in the hybrid framework which covers every key. If I'm not mistaken, only existing keys in the underlying module config can be overwritten by the hybrid config.

The following changes should be considered in the hybrid handler. Since these are quite a few and some a little bit more complex, I think it makes sense to split the work in two stages.
1. Remove obsolete config keys and add missing ones
    - [x] Remove obsolete config keys `Nbins`, `q_max`, `rescatter`, and `weakContribution` (these are not in use since before the first tag `SMASH-hadron-sampler-1.0` and hence can be safely removed from the hybrid framework)
    - [x] Add all possible sampler config keys to hadron sampler config
2. Implement usage of command line syntax according to sampler version (as mentioned by Alessandro down below). For the new version it has to be considered that:
    - [x] the config key `spectra_dir` was renamed to `output_dir`
    - [x] the config key `surface` was renamed to `surface_file`
    - [x] the config key `create_root_output` was added
    - [x] the command line parameters to run the sampler have been changed
    - ~~the prerequisites documentation of the handler should be updated~~ (was not needed after all)

RobinSattler:
[...]

PR #57, opened 2025-02-26
Remove obsolete smash sampler config keys and improve sampler unit test functions

RobinSattler:
This is the fix for the first stage of #52.

**What I did:**
- Remove obsolete config keys `Nbins`, `q_max`, `rescatter`, and `weakContribution` (these are not in use since before the first tag `SMASH-hadron-sampler-1.0` and hence can be safely removed from the hybrid framework).
- Add all missing sampler config keys to hadron sampler config (set to their default values).
- Renamed the unit tests relating to the SMASH-Hadron-Sampler since this was imprecise/confusing when running the tests. The FIST tests have a "-FIST" suffix, so I added a "-SMASH" suffix for the hadron sampler tests.
- Fixed some inconsistencies in the unit test functions

@NGoetz and @AxelKrypton, since this is my first (real) contribution to the handler, I'm looking forward to your feedback.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -7,6 +7,7 @@
 | Niklas Götz        | goetz@itp.uni-frankfurt.de       |
 | Nils Saß           | nsass@itp.uni-frankfurt.de       |
 | Renan Hirayama     | hirayama@itp.uni-frankfurt.de    |
+| Robin Sattler      | sattler@itp.uni-frankfurt.de     |
 | Zuzana Paulinyova  | zuzana.paulinyova@upjs.sk        |
 
 ## Past developers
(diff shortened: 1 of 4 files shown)


NGoetz (approved):
It's only minor changes and it looks good to me.

AxelKrypton (approved):
I am totally happy with these changes. Not even a minor comment. 😉 

Actually I have only a general thought. I implemented the test framework clean-up function in the very beginning as it is a standard feature of testing frameworks. Later on we slightly changed approach and every test is run in a dedicated folder. Since tests folders are deleted on success (unless otherwise requested) and kept on failure, I am wondering which is the point of deleting files in the clean-up function. 🤔 It feels to me like increasing maintenance effort without benefit. @NGoetz Do you see a concrete use case for this? Especially because if the test succeeds and you ran them with the `-k` option, then the dedicated folder would stay, but the files deleted by the clean-up function would still be deleted...

PR #59, opened 2025-03-24
New hybrid logo

AxelKrypton:
@NGoetz @nilssass and I have generated a new in our opinions _way_ better looking logo for the project. As QM is coming up soon, I believe it would be nice to have this in before. We might also think about making a `2.1` release, such that the new logo is also correctly displayed in the README file.

If somebody believes that we should not change the logo at all, let's discuss. I have the feeling the project is not yet super known around, especially since it did not go through a QM conference, yet.


RobinSattler:
I'd be fine with a change, but have two questions:
1. Is it on purpose that the stages do not have equal proportions w.r.t. sections of the circle?
2. If I go around the circle starting at the IC (whirlwind) then the order of the stages is not consecutively in terms of software run in the handler. Is this intended?

AxelKrypton:
Excellent points @RobinSattler, I forgot to mention these.

> 1. Is it on purpose that the stages do not have equal proportions w.r.t. sections of the circle?

Yes. I do not think it is important. And I personally believe that stretching the water and earth elements makes the logo more peaceful and zen in some sense. 😅 

> 2. If I go around the circle starting at the IC (whirlwind) then the order of the stages is not consecutively in terms of software run in the handler. Is this intended?

We got the base of the logo from an AI and it was not possible to make it reordering in a satisfactory way. Given the zen-asymmetry at 1., then one could simply be happy with this. 🙃

RenHirayama:
I will not review this as I do not agree with using AI for such purposes.

PR #61, opened 2025-03-27
Add version logic to handle breaking changes in SMASH-hadron-sampler execution and config key names

### LINKED ISSUE #52: Adjust smash-hadron-sampler execution because of breaking changes in sampler repository

RobinSattler:
This fixes #52.

@AxelKrypton and I added version logic to read out the Sampler version (for the SMASH-hadron-sampler). This is needed to use the appropriate command line parameters to execute the sampler since they were altered from version `3.1.1` to `3.2` resulting in a breaking change. Additionally, two config keys were renamed to clarify them (`surface` to `surface_file`, `spectra_dir` to `output_dir`) which is also considered in our changes and a new sampler base config was added (for version 3.2 and higher) and the old one was renamed.

We also refactored a bit of the SMASH sampler unit testing code and @AxelKrypton did the same to the FIST part of the unit testing on his own.

Be aware that we changed the "test executable" from using the built-in `echo` to a custom `echo` command defined in the `echo.py` file. This ensures that we can set the sampler version for the testing/mocking in a controllable way.

> [!IMPORTANT]
> @NGoetz, it might be smart to start the review with the file _[docs/developer/testing_framework.md](https://github.com/smash-transport/smash-vhlle-hybrid/pull/61/files#diff-64ac464ae73bd3465a090a60d662f451808f28960dbb56c77a1ac0d7c8e53d12)_ since the addition in there clarifies some of the choices (and a potential trap) in the test structures.


bash/Sampler_functionality_SMASH.bash
```diff
@@ -16,36 +16,42 @@ function Create_Superfluous_Symbolic_Link_To_External_Files_Ensuring_Their_Exist
...
         'cs2'
         'ratio_pressure_energydensity'
     )
+    if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
+        allowed_keys+=('create_root_output')
+    fi
+    readonly allowed_keys
     local keys_to_be_found
-    keys_to_be_found=2
+    keys_to_be_found=4
```
NGoetz, line 54: I am not sure if this is ideal. I think the `keys_to_be_found` where 2 because two of the default values in the default config were to be always replaced (`=DEFAULT=`), whereas the other could use the default values. This did not change, so why do we have now 4 instead 2?

NGoetz (commented):
In general the changes make sense to me and effectivley deal with multiple version behavior, and reduce code duplication. I am not sure however if I understand why we have now two more mandatory keys for the sampler. I feel like we mix two concepts - keys which are mandatory to be set for a handler run of the sampler, either through an earlier stage or by hand, and keys which have to be given to the sampler for it to run at all.

bash/Sampler_functionality_SMASH.bash
NGoetz: This is true, but on the other hand our default config always sets these values (other than the paths). On the other hand, the user could define a different default config, and then this change makes sense.
On the other hand, I wonder, how much are we safe against changes of the default config, in the sense, how much manipulation of it can we tolerate. Does the replace placeholder (`=DEFAULT=`) have to be literally this?
So basically, I wonder if `keys_to_be_found` encodes the config keys the sampler does not intialise, or the handler needs to successfully run the sampler. I am fine with both, I just think I understood it differently before.

AxelKrypton: `=DEFAULT=` must be literally so. We do explain this describing [the Sampler section](https://smash-transport.github.io/smash-vhlle-hybrid/develop/user/configuration_file/#the-hadron-sampler-section) of the handler configuration in the user guide.

> So basically, I wonder if `keys_to_be_found` encodes the config keys the sampler does not intialise, or the handler needs to successfully run the sampler. 

I am not sure to see a difference. The Bash variable `keys_to_be_found` denotes how many keys must be present in the sampler config file for the hybrid handler to accept it. The validation itself defines then which these keys are. If the sampler itself abort if a key was not given or leaves the corresponding C++ variable uninitialised, this does not make a different on the hybrid handler side, which consider both cases as a mandatory key to have in the configuration file.

NGoetz: I have been thinking more about his and I think I agree with the change. I don't think it is strictly necessary, but it is another safety measure.

tests/mocks/sampler_black-box.py
```diff
@@ -29,12 +32,23 @@
...
+# behavior which is mimicking the different real sampler versions.
 
 import sys
 import os
 import time
 import random
 import numpy as np
+from packaging.version import Version
+
+sampler_version = os.environ.get("MOCK_HADRON_SAMPLER_VERSION", "3.1.1")
```
NGoetz, line 46: Does it make sense to have as the default the old version, as in the future, we will have the new call?

bash/global_variables.bash
NGoetz, line 1: The copyright needs to be updated

tests/mocks/sampler_black-box.py
AxelKrypton: Ideally, I would have preferred not to have a default one. Just to say: The test must decide how to test. But I did not find time for that and I also think that it is not so bad to have the old one. As you say, we will more and more often use the newer versions. However, we do not set any requirement, so it is not bad to check by default that everything still works with the older sampler. 🤔

NGoetz (approved):
I went through it a second time and have only minor comments.

tests/mocks/sampler_black-box.py
NGoetz: Sure, it works like this too. Maybe it will make future people appreciate the new way to call the sampler more :)

PR #69, opened 2025-04-25
Implement selection of IC base configuration files depending on software version

### LINKED ISSUE #66: SMASH-3.2 key was added to base config which leads to "support" only of latest SMASH version
In commit daa5f83, 

```yaml
        Initial_Conditions:
            Type: "Constant_Tau"
```

was added to the base IC config in the section `Modi: Collider`. Since this key was moved from the `Output` section (SMASH-3.1 and earlier) into the section `Modi: Collider` (for SMASH-3.2 and later) of the smash config in PR smash-transport/smash-devel#1267, this new base config is only supported by the latest SMASH version `3.2`. The user could always alter this base config or use custom base configs, but the status quo is not very user friendly.

To be backward compatible and user friendly, we should introduce an old and a new base config from which the right one will be picked using version logic inside the handler framework, similar to the handling of the sampler base configs (see #61 for more insights).

AxelKrypton:
This closes #66.

I implemented a similar mechanism as it was recently done for the Sampler stage. In this case it was slightly easier as for IC we support only SMASH. I also assumed that `--version` has always been there in SMASH in the sense that if this was not the case a generic error is printed.

If the user is using the development version of SMASH like version `SMASH-3.2-18-gc4c0322a4`, the base config file is that of version `3.2`.

@RobinSattler I did not try a real run yet, maybe you can start with that. Also Git displays the changes in the sanity checks in a weird way as if I changed code. Actually I just added the case for `IC`, but adding code before the case for the `Sampler` made Git interpreting this so.


tests/unit_tests_IC_functionality.bash
```diff
@@ -22,17 +22,55 @@ function Make_Test_Preliminary_Operations__IC-create-input-file()
...
     # Touch dummy empty handler config as this is always there in sanity checks
     touch "${HYBRID_configuration_file}"
+}
+
+function Make_Test_Preliminary_Operations__IC-pick-correct-base-config()
+{
+    __static__Do_Preliminary_IC_Setup_Operations
+}
+
+function __static__Is_Picked_IC_Base_Config_For_Version()
```
RobinSattler, line 36: I think the function name is not really self-explanatory. What about `__static__Is_Picked_IC_Base_Config_Compatible_With_Version` or `__static__Is_Right_IC_Base_Config_Picked_For_Version`?

tests/unit_tests_IC_functionality.bash
```diff
@@ -22,17 +22,55 @@ function Make_Test_Preliminary_Operations__IC-create-input-file()
...
+    :
+}
+
+function Make_Test_Preliminary_Operations__IC-create-input-file()
+{
+    __static__Do_Preliminary_IC_Setup_Operations
+    export MOCK_ECHO_VERSION=3.1
     Perform_Sanity_Checks_On_Provided_Input_And_Define_Auxiliary_Global_Variables
+    # Since we use our mock of echo as fake sampler executable the function above will set the
+    # sampler version to the MOCK_ECHO_VERSION environment variable value.
```
RobinSattler, line 69: Shouldn't that be `IC` instead of `sampler` in this comment?

tests/unit_tests_IC_functionality.bash
```diff
@@ -22,17 +22,55 @@ function Make_Test_Preliminary_Operations__IC-create-input-file()
...
 }
 
 function Unit_Test__IC-create-input-file()
 {
-    touch "${HYBRID_software_base_config_file[IC]}"
```
RobinSattler, line 35: Just for my understanding: since we call function `__static__Is_Picked_IC_Base_Config_For_Version` before this function, the IC base config is already picked and set and hence we do not need to create a dummy config anymore?

RobinSattler (commented):
I ran the IC part using the hybrid handler with SMASH-3.1 and SMASH-3.2, both worked flawlessly. Additionally, I ran the whole evolution for SMASH-3.2 without an error*. The majority of the changes looks good, but I have a couple of questions (see comments in the code) which might be worth discussing before I approve.

*I just noticed that I get a warning running the full evolution (all stages) - see screenshot down below - because I changed the shear value in the base config from `shear 1` to `shear 0` in PR #57 and this is not consistent with the hydro base config. Could you set `shear` back to `1` in both smash-hadron-sampler base configs?

Screenshot 2025-04-26 at 22 55 26

Additionally, in _sanity_checks.bash_ in line 266 (your branch): is there a reason why this is not read-only? I don't see any so it might be good to add the `-r`. 🙂

bash/sanity_checks.bash
```diff
@@ -57,9 +57,9 @@ function Perform_Internal_Sanity_Checks()
         'These Python scripts should be shipped within the hybrid handler codebase.' '--' \
         "${HYBRID_external_python_scripts[@]}"
     for key in "${!HYBRID_software_base_config_file[@]}"; do
```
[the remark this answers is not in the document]
AxelKrypton: Actually not. `HYBRID_software_base_config_file` is an associative array and hence `"${!HYBRID_software_base_config_file[@]}"` expands to a properly quoted list of the keys of the array. That's what we want here. Excluding the `IC` and `Sampler` keys are needed, as those entries do not have a filename as value yet at this point.

bash/sanity_checks.bash
```diff
@@ -307,40 +325,66 @@ function __static__Set_Software_Input_Data_File_If_Not_Set_By_User()
...
-        # sampler versions exit with code 1 if a command line option is not recognized. Hence, we can
-        # confidently give an error and fail if the sampler fails with an exit code different from 1.
-        #
-        # NOTE: Being errexit option enabled, we need to store the version output in the if-clause
-        #       condition and access the possible exit code at the very beginning of the else-clause.
-        if sampler_version_output=$(${HYBRID_software_executable[Sampler]} --version 2> /dev/null); then
-            if [[ ${sampler_version_output} =~ ${HYBRID_version_regex} ]]; then
-                HYBRID_software_version[Sampler]="${BASH_REMATCH[0]}"
-                Print_Debug 'Sampler version found to be ' --emph "${HYBRID_software_version[Sampler]}" '.'
+    case "$1" in
```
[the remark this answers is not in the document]
AxelKrypton: The reason was that it was on the very first line of the function, but it is fine to use a `local -r` variable, too. I changed it now.

tests/unit_tests_IC_functionality.bash
AxelKrypton: True. I chose `__static__Is_Picked_IC_Base_Config_Correct_For_Version`. 😉

tests/unit_tests_IC_functionality.bash
AxelKrypton: Before this PR the unit test of `Prepare_Software_Input_File_IC` was using a dummy file was base config. As we are unit testing the preparation of the config, e.g. the copy, we are not interested in testing what is copied. This was the rationale before. Now I let also this unit test use the base config file, whatever it is. It would not be wrong to touch it, but unnecessary (as it is an existent file now). However, it would be wrong to remove it in the test clean-up, which I did by accident during development and I had a base config file disappearing anytime I was running this test... 🫠

RobinSattler (approved):
Just a tiny thing to fix, otherwise it looks good and I already approve! Thanks for the changes! :rocket:

Issue #68, opened 2025-04-24
Implement sanity check for run with all hybrid stages and `Add_spectators_from_IC: true`


RobinSattler:
After discussing how to proceed with fixing the _add_spectators.py_ file (issue #65), @AxelKrypton, @RenHirayama, and I decided to restrict the hybrid handler to only one IC event runable if the `Add_spectators_from_IC: true` key is set in the Afterburner section. To ensure this, the following check should be added to the _bash/sanity_checks.bash_ file:
If IC and Afterburner stage (so full evolution) are given in the config(s) as well as the `Add_spectators_from_IC: true`, the run should be aborted with an error if `Nevents` in the IC config is larger than 1.

The reason for this is that right now the python file to add the spectators can't handle multiple events in the IC and we'll leave this as work for the future (see issue #67).

RobinSattler:
After the recent development in issue #67 and the possible fix for it, this might not be needed after all.

AxelKrypton:
I would say that the hybrid handler should still check the number of events used in `IC` and `Sampler` phases and warn the user if these are different.

RobinSattler:
Good point! As far as I can see, this is not yet implemented - contrary to what we thought earlier. Such a check is only implemented between the sampler and afterburner events. Appartently, the afterburner config can take `Nevents` even though smash is run in list modus. 🤔

PR #71, opened 2025-05-01
Fix bugs in `add_spectators.py` script and restrict possibility of adding spectators to only one IC event

### LINKED ISSUE #65: `add_spectators.py` python script contains multiple problems
I noticed a couple of issues in the _python/add_spectators.py_ file:

1. Extracts the projectile particles twice from the config instead of once projectile and once target particles (see [lines 36-37](https://github.com/smash-transport/smash-vhlle-hybrid/blob/a94a98333ae44bfceab3e3929766856b0bd19e44/python/add_spectators.py#L36-L37))
2. Due to the incorporation of the ensemble number in the event headers the if statement in [line 95](https://github.com/smash-transport/smash-vhlle-hybrid/blob/a94a98333ae44bfceab3e3929766856b0bd19e44/python/add_spectators.py#L95) does not work anymore for SMASH-3.2
3. The calling of function `get_initial_nucleons_from_config()` in [line 131](https://github.com/smash-transport/smash-vhlle-hybrid/blob/a94a98333ae44bfceab3e3929766856b0bd19e44/python/add_spectators.py#L131) is unnecessary (not stored anyway and the proper call is in [line 59](https://github.com/smash-transport/smash-vhlle-hybrid/blob/a94a98333ae44bfceab3e3929766856b0bd19e44/python/add_spectators.py#L59))

There might be more problems, but I have to check.
@AxelKrypton, should this (or at least item 2) be a bug fix for `2.1` (i.e., should I branch off of `main` to fix it)?

RobinSattler:
This fixes #65.

- I fixed the `add_spectators.py` script and it should now work SMASH version independent.
- Additionally, the handler will now print an error and exit when `Nevents` in the IC is not one and `Add_spectators_from_IC: true` is given in the afterburner config. It's important to state that this error and exit is only done in the afterburner stage so far, as soon as the python script `add_spectators.py` is called. The change to "sanity" check before the handler runs if more than one IC event is run combined with `Add_spectators_from_IC: true` in the afterburner section will be left for issue #68.
- On top of our testing framework, I did test runs with SMASH-3.1 and SMASH-3.2 (and the corresponding SMASH-hadron-sampler versions) and everything went as expected. I tested also SMASH-3.1 with the old script and with the new implementation and with the same random seed there is (basically) no difference, which is what we would expect for a symmetric collision (same nuclei as target and projectile).

One point up for discussion:
- Since PR #70 renames `sampled_particles_list.oscar` to `sampled_particles.oscar`, should we rename the command line option of the python script as well (e.g., from `--sampled_particle_list` to `--sampled_particles_file`)?

@RenHirayama, since you said that adding spectators is relevant for your research, I assign you as well to review.

### DIFF
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -1,6 +1,6 @@
 #===================================================
 #
-#    Copyright (c) 2023-2024
+#    Copyright (c) 2023-2025
 #      SMASH Hybrid Team
 #
 #    GNU General Public License (GPLv3 or later)
@@ -77,6 +77,19 @@ function __static__Ensure_Consistency_Of_Afterburner_Input()
     fi
 }
 
+function __static__Run_Add_Spectators_Python_Script()
+{
+    local python_fatal_value_error
+    # NOTE: To extract the right error code in the python script, it has to be handed to it right
+    #       before calling the script since the error codes are not exported by default.
+    python_fatal_value_error=${HYBRID_fatal_value_error} \
+        "${HYBRID_external_python_scripts[Add_spectators_from_IC]}" \
+        '--sampled_particle_list' "${HYBRID_software_input_file[Afterburner]}" \
+        '--initial_particle_list' "${HYBRID_software_input_file[Spectators]}" \
+        '--output_file' "${target_link_name}" \
+        '--smash_config' "${HYBRID_software_output_directory[IC]}/config.yaml" 2> /dev/null
+}
+
 function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Without_Spectators()
 {
     local -r target_link_name="${HYBRID_software_output_directory[Afterburner]}/${HYBRID_afterburner_list_filename}"
@@ -92,11 +105,29 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Withou
             "${HYBRID_software_input_file[Afterburner]}" \
             "${HYBRID_software_input_file[Spectators]}"
         # Run Python script to add spectators
-        "${HYBRID_external_python_scripts[Add_spectators_from_IC]}" \
-            '--sampled_particle_list' "${HYBRID_software_input_file[Afterburner]}" \
-            '--initial_particle_list' "${HYBRID_software_input_file[Spectators]}" \
-            '--output_file' "${target_link_name}" \
-            '--smash_config' "${HYBRID_software_output_directory[IC]}/config.yaml"
+        # NOTE: Since the errexit option enabled, the python script to add the spectators is run in the if-statement
+        #       and the possible exit code is accessed at the very beginning of the else-clause.
+        local python_exit_code
+        if __static__Run_Add_Spectators_Python_Script; then
+            # BE AWARE: The negation of the if-clause does not work because then the exit code cannot
+            #           be extracted properly (it will always be 0 because of the true if-statement).
+            :
+        else
+            python_exit_code=$?
+            if [[ ${python_exit_code} -eq ${HYBRID_fatal_value_error} ]]; then
+                exit_code=${HYBRID_fatal_value_error} Print_Fatal_And_Exit \
+                    'It was attempted to add spectators from multiple IC events to the sampled particles file. Only' \
+                    'running one IC event is supported when using the Afterburner config key ' \
+                    --emph "Add_spectators_from_IC: true" '.'
+            elif [[ ${python_exit_code} -eq 2 ]]; then
+                Print_Internal_And_Exit \
+                    'The handing over of the ' --emph "python_fatal_value_error" \
+                    ' to the Python script that adds spectators from the IC did not work.'
+            else
+                Print_Fatal_And_Exit \
+                    'Adding spectators from the IC particles to the sampled particles file failed.'
+            fi
+        fi
     else
         if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
             ln -s -f "${HYBRID_software_input_file[Afterburner]}" "${target_link_name}"
--- a/python/add_spectators.py
+++ b/python/add_spectators.py
@@ -2,7 +2,7 @@
 
 #===================================================
 #
-#    Copyright (c) 2023-2024
+#    Copyright (c) 2023-2025
 #      SMASH Hybrid Team
 #
 #    GNU General Public License (GPLv3 or later)
@@ -29,12 +29,12 @@ def get_initial_nucleons_from_config():
         nuclei and read off the number of constituents.
     '''
     N_initial_nucleons = 0
-    with open(args.smash_config, 'r') as file:
-        data_config = yaml.safe_load(file)
+    with open(args.smash_config, 'r') as config_file:
+        data_config = yaml.safe_load(config_file)
 
     try:
         projectile_particles = data_config['Modi']['Collider']['Projectile']['Particles']
-        target_particles = data_config['Modi']['Collider']['Projectile']['Particles']
+        target_particles = data_config['Modi']['Collider']['Target']['Particles']
 
         for particle in projectile_particles.keys():
             N_initial_nucleons += projectile_particles[particle]
@@ -42,8 +42,9 @@ def get_initial_nucleons_from_config():
         for particle in target_particles.keys():
             N_initial_nucleons += target_particles[particle]
     except:
-        print("The config file does not contain the necessary information about the projectile and target nuclei.")
-        sys.exit()
+        print("The config file does not contain the necessary information "
+              "about the projectile and/or target nuclei.")
+        sys.exit(1)
 
     return N_initial_nucleons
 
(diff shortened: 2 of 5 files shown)


python/add_spectators.py
```diff
@@ -63,7 +70,9 @@ def extract_spectators():
         for line in f:
             if line[0] == "#": continue  # Comment line
             # Is initial nucleon and has not interacted
-            if ((int(line.split()[10]) <= N_nucleons) and (int(line.split()[12]) == 0) ):
+            particle_id = int(line.split()[10])
```
RenHirayama, line 73: I don't think we need this condition, as if the particle has Ncoll==0 it is an initial hadron (particles created during the evolution start with Ncoll==1). But I guess it doesn't hurt

python/add_spectators.py
```diff
@@ -89,15 +98,15 @@ def write_full_particle_list(spectator_list):
...
+        for line in sampler_output_file:
             # Find header of each new event
-            if (line.startswith('# event') and len(line.split()) == 5 ):
-                # 5th column contains the number of sampled particles for each
-                # event. To this value the number of spectators needs to be added
-                particles_in_event = int(line.split()[4]) + N_spectators
+            if (line.startswith('# event') and (not "end" in line)):
+                # The last column contains the number of sampled particles for each
+                # event. The number of spectators needs to be added to this value
+                particles_in_event = int(line.split()[-1]) + N_spectators
```
RenHirayama, line 107: In the interest of being safe we should instead look for the string 'out' and get the next string

RenHirayama (approved):
Looks ok to me. I think the way we extract spectators could be better from the smash side, but that's beyond this PR, and since my comments are not major I will approve already

python/add_spectators.py
RobinSattler: Probably not very likely that this happens a lot, but isn't there a chance that initial nucleons end up having just elastic collisions? Then they would be participants and still have their old `ID`. I think it's safer to keep it this way.

RenHirayama: I meant that the check on `N_collisions` should be enough to determine and we don't need the ID. But we can keep the check, it doesn't create any problem.

And yes, some nucleons will likely only interact elastically (~25% at 10 GeV), but they are fed to vHLLE afaik.

RobinSattler: Ah, I thought you meant the other way around. 😄 

Interesting, thanks for sharing!

bash/Afterburner_functionality.bash
```diff
@@ -92,11 +92,22 @@ function __static__Create_Sampled_Particles_List_File_Or_Symbolic_Link_With_Or_W
             "${HYBRID_software_input_file[Afterburner]}" \
             "${HYBRID_software_input_file[Spectators]}"
         # Run Python script to add spectators
+        exit_code=0
```
AxelKrypton, line 95: I would (1) force this from within the Python script adding there at the and an exit with `0` code (which might well be the default, to be checked) and (2) avoid using the global logger variable. I know it should not be a problem, but it is kind of according to the "let's avoid future headaches" philosophy. 😅 Using a `local python_exit_code` variable is also more specific.

bash/Afterburner_functionality.bash
```diff
@@ -92,11 +92,22 @@ function __static__Create_Sampled_Particles_List_File_Or_Symbolic_Link_With_Or_W
...
         "${HYBRID_external_python_scripts[Add_spectators_from_IC]}" \
             '--sampled_particle_list' "${HYBRID_software_input_file[Afterburner]}" \
             '--initial_particle_list' "${HYBRID_software_input_file[Spectators]}" \
             '--output_file' "${target_link_name}" \
-            '--smash_config' "${HYBRID_software_output_directory[IC]}/config.yaml"
+            '--smash_config' "${HYBRID_software_output_directory[IC]}/config.yaml" \
+            || exit_code=$?
+        if [[ ${exit_code} -eq ${HYBRID_fatal_value_error} ]]; then
```
AxelKrypton, line 102: I leave only here this comment which is also about new line 39 of the Python script. If you exit with `68` in Python and here use the global hybrid variable storing the error code, you add a sneaky dependency. Not that I suggest that this should happen in the future, but changing the number contained in `HYBRID_fatal_value_error` would break the intended behaviour of your code. And having the global variables to store exit codes is in principle meant to allow changing the exit codes without problems... 😅

You can either take the value from the environment from within Python (we do this already e.g. in the _echo.py_ script) - and you might consider failing if that it is not set - or you simply hard-code the exit code ob both sides. This is also fine if we consider the Python script more disconnected from the hybrid handler.

If you go for extracting from environment and fail if the env var is not set, then I would exit e.g. `2` there and handle this exiting the handler with an internal error.

AxelKrypton (approved):
@RobinSattler I also approve, excellent work and important fixes.

I have one Bash/clean code point that I hope I explained well enough. Ask if it is unclear. I think it is worth adjusting this.

In the sprint we will merge this and the other PR once decided which direction to take w.r.t. next release.

bash/Afterburner_functionality.bash
RobinSattler: Good point. I like the readout from the environment. For this, I'd have to `export` the error code in _bash/error_codes.bash_. Is there a drawback to this? Should I do this with all of them then?

bash/Afterburner_functionality.bash
RobinSattler: > (1) force this from within the Python script adding there at the and an exit with 0 code

If I don't set `local python_exit_code=0` beforehand, is there a smart way to extract the exit code?
The only idea I have so far is to use an auxiliary variable for it:
```bash
        local python_exit_code aux_exit_code=''
        "${HYBRID_external_python_scripts[Add_spectators_from_IC]}" \
            '--sampled_particle_list' "${HYBRID_software_input_file[Afterburner]}" \
            '--initial_particle_list' "${HYBRID_software_input_file[Spectators]}" \
            '--output_file' "${target_link_name}" \
            '--smash_config' "${HYBRID_software_output_directory[IC]}/config.yaml" \
            || aux_exit_code=$?
        python_exit_code=$?
        if [[ "${aux_exit_code}" != '' ]]; then
            python_exit_code="${aux_exit_code}"
        fi
```

AxelKrypton: Working with `errexit` option enabled, usually you put the command in the `if`-clause, possibly delegating it to a function if it is complex.

In this simple case, given that [it seems that Python returns 0 on success](https://stackoverflow.com/questions/36136480/what-is-pythons-default-exit-code), your previous approach is fair enough. I mean, still with a `local python_exit_code=0` and then `|| python_exit_code=$?`. After all this was a nuance; my main concern was more avoid using the logger global to stay away from possible sneaky dependencies (logger is a separate independently developed project, who knows what will happen there... 😉).

RobinSattler: Thanks for your input! 🙂 I went for the simpler change and will leave this unresolved for now since we got a little time left until we'll merge this PR. In the meantime, we (or I) can think about implementing the standard `if`-statement option.

RobinSattler: Thanks for our discussion in person. I implemented it after all. :muscle:

bash/Afterburner_functionality.bash
```diff
@@ -77,6 +77,19 @@ function __static__Ensure_Consistency_Of_Afterburner_Input()
     fi
 }
 
+function __static__Run_Add_Spectators_Python_Script()
+{
+    local python_fatal_value_error
```
AxelKrypton, line 82: Strictly speaking this variable is not needed to be declared as local, since the chance on line 85 where it is set is only visible from the command run in that line, here the python script. Feel free to remove it or leave it, if it feels better, or remove it and add a comment why this variable is not needed to be declared local. 😉

bash/Afterburner_functionality.bash
```diff
@@ -92,11 +105,28 @@ function __static__Create_Sampled_Particles_List_File_Or_Symbolic_Link_With_Or_W
...
+        else
+            python_exit_code=$?
+            if [[ ${python_exit_code} -eq ${HYBRID_fatal_value_error} ]]; then
+                exit_code=${HYBRID_fatal_value_error} Print_Fatal_And_Exit \
+                    'It was attempted to add spectators from multiple IC events to the sampled particles file. Only' \
+                    'running one IC event is supported when using the Afterburner config key ' \
+                    --emph "Add_spectators_from_IC: true" '.'
+            elif [[ ${python_exit_code} -eq 2 ]]; then
+                Print_Internal_And_Exit \
+                    'The Hybrid-handler variable ' --emph "HYBRID_fatal_value_error" ' does not seem to be set.'
```
AxelKrypton, line 124: I think here we should point to the variable that the Python script tries to use.

```suggestion
                Print_Internal_And_Exit \
                    'The handing over of the ' --emph "python_fatal_value_error" \
                    ' to the Python script to add spectators did not work.'
```

AxelKrypton (approved):
Excellent additions. 🙌  I left another couple of minor comments.

bash/Afterburner_functionality.bash
RobinSattler: I'm in favor of leaving it as it is, but thanks for pointing it out!

PR #74, opened 2025-05-15
Add custom inputfile name

RenHirayama:
This changes how the `Input_file` under `Hydro` and `Afterburner` behave. Now one can give a custom input file name, and the handler will look for this file name in the appropriate sub-folders, instead of the default _SMASH\_IC.dat_ and _particle_lists.oscar_ files in `IC` and `Sampler`


bash/sanity_checks.bash
```diff
@@ -299,12 +299,16 @@ function __static__Set_Software_Input_Data_File_If_Not_Set_By_User()
...
+        elif [[ "${filename}" =~ / ]]; then
             if Element_In_Array_Equals_To "${relative_key}" "${HYBRID_given_software_sections[@]}"; then
                 exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
                     'Requesting custom ' --emph "${key}" ' input file although executing ' \
                     --emph "${relative_key}" ' with default output name.'
             fi
+        else
+            printf -v filename '%s/%s' \
+                "${HYBRID_software_output_directory[${relative_key}]}" \
+                "${filename}"
```
AxelKrypton: Since I am not totally sure that this input is validated at any point, we could probably catch here basic misuse. In general, the Hybrid handler at some points calls
```bash
Ensure_Given_Files_Exist "${HYBRID_software_input_file[...]}"
```
and this will fail if e.g. the provided input file is like a directory. At the moment the user can provide `.` or `..` as input file. This will fall into this `else` and I am wondering whether we should avoid that... 🤔

docs/user/CHANGELOG/index.md
```diff
@@ -48,6 +48,8 @@ Given a version number `X.Y.Z`,
 
     **Changes:**
 
+    - `Input_file` in the `Hydro` and `Afterburner` modules now can be used to modify the expected file names from the appropriate :file_folder: ***IC*** and :file_folder: ***Sampler*** sub-folders, when a string without a "/" character is given.
```
AxelKrypton: You should pick an emoji instead of a bullet point. Used ones are just listed above. I guess `:new:` is appropriate here. I would also use `` `/` `` instead of `"/"`.

```suggestion
    :new: The `Input_file` configuration key in the `Hydro` and `Afterburner` modules can now be used to modify the expected file names from the appropriate :file_folder: ***IC*** and :file_folder: ***Sampler*** sub-folders, when a string without a `/` character is given.
```

docs/user/configuration_file.md
```diff
@@ -141,7 +141,8 @@ IC:
 
     The hydrodynamics simulation needs an additional input file which contains the system's initial conditions.
     This is the main output of the previous stage and, therefore, if not specified, a :material-file: *SMASH_IC.dat* file is expected to exist in the :file_folder: ***IC*** output sub-folder with the same `Run_ID`.
-    However, using this key, any file can be specified and used.
+
+    However, using this key, any file can be specified and used. If the key is a simple file name (without any '/'), the hybrid handler looks for this name in the corresponding :file_folder: ***IC*** output sub-folder, but if it is a specific path (containing '/'), that specific file will be used.
```
AxelKrypton: Keep a sentence per line.

```suggestion
    However, using this key, any file can be specified and used.
    If the key is a simple file name (without any `/`), the hybrid handler looks for this name in the corresponding :file_folder: ***IC*** output sub-folder, but if it is a path (i.e. it contains a `/`), that specific file will be used.
```

docs/user/configuration_file.md
```diff
@@ -208,7 +209,9 @@ Sampler:
 
     As other stages, the afterburner run needs an additional input file as well, one which contains the sampled particles list.
     This is the main output of the previous sampler stage and, therefore, if not specified, a *particle_lists.oscar* file is expected to exist in the ***Sampler*** output sub-folder with the same `Run_ID`.
-    However, using this key, any file can be specified and used.
+
+    However, using this key, any file can be specified and used. If the key is a simple file name (without any '/'), the hybrid handler looks for this name in the corresponding :file_folder: ***Sampler*** output sub-folder, but if it is a specific path (containing a '/'), that specific file will be used.
```
AxelKrypton: Similar suggested changes here.

```suggestion
    However, using this key, any file can be specified and used.
    If the key is a simple file name (without any `/`), the hybrid handler looks for this name in the corresponding :file_folder: ***Sampler*** output sub-folder, but if it is a path (i.e. it contains a `/`), that specific file will be used.
```

tests/functional_tests_Afterburner_only.bash
```diff
@@ -43,6 +43,26 @@ function Functional_Test__do-Afterburner-only()
     Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
     __static__Check_Successful_Handler_Run $?
     mv 'Afterburner' 'Afterburner-success'
+    # Expect success with custom input file name from Sampler
+    custom_input="afterburner_input.dat"
```
AxelKrypton: This should be a `local -r` variable.

tests/functional_tests_Afterburner_only.bash
```diff
@@ -43,6 +43,26 @@ function Functional_Test__do-Afterburner-only()
...
+      Add_spectators_from_IC: false
+      Input_file: %s
+    ' "${run_id}" "${HYBRIDT_repository_top_level_path}" "${custom_input}" > "${config_filename}"
+    # Run the hydro stage and check if freezeout is successfully generated
+    Print_Info 'Running Hybrid-handler expecting success'
+    Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
+    if [[ $? -ne 0 ]]; then
+        Print_Error 'Hybrid-handler unexpectedly failed.'
+        return 1
+    fi
```
AxelKrypton: I would simply call

```bash
__static__Check_Successful_Handler_Run $?
```

instead of this manual check.

tests/functional_tests_Hydro_only.bash
```diff
@@ -34,6 +34,26 @@ function Functional_Test__do-Hydro-only()
     fi
     Check_If_Software_Produced_Expected_Output 'Hydro' "$(pwd)/Hydro"
     mv 'Hydro' 'Hydro-success'
+    #Expect success with custom input file name from IC
+    custom_input="hydro_input.dat"
```
AxelKrypton: This should be a `local -r` variable.

AxelKrypton (changes requested):
The changes are fine. I prefer to go once more through them once you dealt with my suggestions. My main extra point here is that, although the functional tests are fine, I would probably think that an integration test is worth being added. Basically calling the config file parsing and the sanity checks and then testing that the output global variable has been properly set. You can check out the only existing integration test to know how to do it and what I mean it. Ask if needed. 😊

RenHirayama:
In my machine the formatting test passes, which I assume it shouldn't given that the code is visibly not formatted. The requirements don't help, everything is up to date apparently. Do you have an idea about this @AxelKrypton?

RobinSattler:
@RenHirayama, I just saw this and checked the formatting test. For me it fails. If I format the codebase, this is the change the formatter does:

Screenshot 2025-05-20 at 20 44 49

RenHirayama:
:shaking_face: turns out that running `./Hybrid-handler format`  does this! (At least in my local computer)

AxelKrypton:
> 🫨 turns out that running `./Hybrid-handler format` does this! (At least in my local computer)

🫠 _It turns out_ that this is mentioned in the [contributing documentation page](https://smash-transport.github.io/smash-vhlle-hybrid/latest/developer/contributing/#bash-notation-in-the-codebase) of the developer guide... 😉

RenHirayama:
I meant that running the formatter *causes* the wrong formatting in my machine. Notice the lack of space between `(` and `#`

image

AxelKrypton:
> I meant that running the formatter _causes_ the wrong formatting in my machine. Notice the lack of space between `(` and `#`

Now I see what you meant by "does this". 🫠 I take back my sentence, sorry. Still, it is strange. Can you tell me which version of `shfmt` are you using? I have `3.7.0` on my machine.

RenHirayama:
`shfmt -version` just returns `(devel)`, but when I try to update it, it says that the newest version is installed (3.4.3-1) :thinking:

bash/sanity_checks.bash
```diff
@@ -299,12 +299,20 @@ function __static__Set_Software_Input_Data_File_If_Not_Set_By_User()
...
             if Element_In_Array_Equals_To "${relative_key}" "${HYBRID_given_software_sections[@]}"; then
                 exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
                     'Requesting custom ' --emph "${key}" ' input file although executing ' \
                     --emph "${relative_key}" ' with default output name.'
             fi
+        else
+            if [[ "${filename}" =~ ^('.'|'..')$ ]]; then
+                exit_code=${HYBRID_fatal_wrong_config_file} Print_Fatal_And_Exit \
+                    'Input_File of ' --emph "${key}" ' cannot be ' "${filename}"
+            fi
```
AxelKrypton: Two remarks:
* In Bash, for mind sake, you do not use quotes in a regex. Your line of code is technically correct, but I would bet most of people would be unsure to the question whether the quote is being matched or not. It is way more common to test this like `=~ ^[.][.]?$ ]]`, namely one literal period followed optionally by another one.
* I would `--emph` the filename as well and add a period afterwards.

```bash
            if [[ "${filename}" =~ ^[.][.]?$ ]]; then
                exit_code=${HYBRID_fatal_wrong_config_file} Print_Fatal_And_Exit \
                    'Input_File of ' --emph "${key}" ' cannot be ' --emph "${filename}" '.'
            fi
```

AxelKrypton (approved):
Great. 🙌 I approve but left a comment to better write the Bash regex. I also took the freedom to improve documentation in the user guide with an annotation. Please have a look to it to check that it makes sense. 💪

AxelKrypton:
> `shfmt -version` just returns `(devel)`, but when I try to update it, it says that the newest version is installed (3.4.3-1) 🤔

I do not understand this. According to [its page](https://github.com/mvdan/sh) you should be able to update it. Last version is 3.11.

PR #75, opened 2025-05-16
Fix IC bug about not considering user custom base config file

AxelKrypton:
This fixes a bug noticed yesterday by @RenHirayama of course few hours after having rolled out the hot-fix `2.1.1`.

@RobinSattler I prepared another hot-fix branch, rebased your branch onto `main` and already bumped the version to mark the beginning of the hot-fix.

About the fix, this was simple and Robin correctly implemented it. I took the opportunity to add a test catching it and this is not a unit test but an integration one. Hence, I added a dedicated mode to the `tests_runner` and make GitHub action run integrations tests, too. There is only one ATM, but I will open an issue as I am sure some "unit" tests are integration ones.

### DIFF
--- a/tests/integration_tests_custom_base_configuration.bash
+++ b/tests/integration_tests_custom_base_configuration.bash
@@ -21,7 +21,7 @@ function Make_Test_Preliminary_Operations__set-base-config-file-from-user()
     Define_Further_Global_Variables
 }
 
-function Unit_Test__set-base-config-file-from-user()
+function Integration_Test__set-base-config-file-from-user()
 {
     local -r functions_to_call=(
         'Validate_And_Parse_Configuration_File'
--- a/tests/integration_tests_custom_base_configuration.bash
+++ b/tests/integration_tests_custom_base_configuration.bash
@@ -7,7 +7,7 @@
 #
 #===================================================
 
-function Make_Test_Preliminary_Operations__set-base-config-file-from-user()
+function Make_Test_Preliminary_Operations__set-custom-base-config-file()
 {
     local file_to_be_sourced list_of_files
     list_of_files=(
@@ -21,7 +21,7 @@ function Make_Test_Preliminary_Operations__set-base-config-file-from-user()
     Define_Further_Global_Variables
 }
 
-function Integration_Test__set-base-config-file-from-user()
+function Integration_Test__set-custom-base-config-file()
 {
     local -r functions_to_call=(
         'Validate_And_Parse_Configuration_File'
--- a/.github/workflows/github-actions-on-github-servers.yml
+++ b/.github/workflows/github-actions-on-github-servers.yml
@@ -13,6 +13,8 @@ on:
     branches:
       - main
       - develop
+      - release*
+      - hotfix*/*
 
 jobs:
   run_tests:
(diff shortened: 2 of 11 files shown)


tests/integration_tests_custom_base_configuration.bash
```diff
@@ -0,0 +1,63 @@
+#===================================================
+#
+#    Copyright (c) 2025
+#      SMASH Hybrid Team
+#
+#    GNU General Public License (GPLv3 or later)
+#
+#===================================================
+
+function Make_Test_Preliminary_Operations__set-base-config-file-from-user()
```
RobinSattler, line 10: I think I'd prefer something like this:
```suggestion
function Make_Test_Preliminary_Operations__set-custom-base-config-file()
```

tests/integration_tests_custom_base_configuration.bash
```diff
@@ -0,0 +1,63 @@
...
+        'global_variables.bash'
+        'sanity_checks.bash'
+    )
+    for file_to_be_sourced in "${list_of_files[@]}"; do
+        source "${HYBRIDT_repository_top_level_path}/bash/${file_to_be_sourced}" || exit ${HYBRID_fatal_builtin}
+    done
+    Define_Further_Global_Variables
+}
+
+function Integration_Test__set-base-config-file-from-user()
```
RobinSattler, line 24: Same here:
```suggestion
function Integration_Test__set-custom-base-config-file()
```

tests/integration_tests_custom_base_configuration.bash
```diff
@@ -0,0 +1,63 @@
...
+                "${HYBRIDT_repository_top_level_path}" > "${HYBRID_configuration_file}"
+            for my_function in "${functions_to_call[@]}"; do
+                Call_Codebase_Function "${my_function}" #&> /dev/null
+            done
+            if [[ "${HYBRID_software_base_config_file[${key}]}" != "${custom_base_config[${key}]}" ]]; then
+                Print_Error 'Wrong base config file ' --emph "${HYBRID_software_base_config_file[${key}]}" \
+                    ' set for ' --emph "${key}" ' stage.'
+                exit 1
+            fi
+        )
```
RobinSattler, line 58: Just for my understanding: this is done in a subshell so `exit 1` can be used without the whole testing framework exiting?

RobinSattler (approved):
I ran the handler with a custom `IC` base config, which worked flawless. Additionally, I tried running all the tests (functional, integration, and unit) in a local repo, again without any problems. 💪

One thing which is - at least to me - a little confusing is the naming of the following files:
`tests/unit_tests_utility_functions.bash`
`tests/utility_functions_unit.bash`
I wouldn't know from the names what the difference between those files is. 🤔 @AxelKrypton, are you aware of this?

Anyway, since I only have minor comments, everything works, and the code changes look sound, I already approve the fix! Thanks for implementing the integration test! 🚀

AxelKrypton:
> One thing which is - at least to me - a little confusing is the naming of the following files:
> `tests/unit_tests_utility_functions.bash`
> `tests/utility_functions_unit.bash`
> I wouldn't know from the names what the difference between those files is. 🤔 @AxelKrypton, are you aware of this?

All files that start with *unit_tests_* are file containing unit tests. The first contains unit tests of the utility functions of the codebase, while the second contains the utility functions of the unit-testing framework and there is an analogous file for the integrations and functional tests.

I will now implement your suggestions and roll out the hot fix.

tests/integration_tests_custom_base_configuration.bash
AxelKrypton: Yes, correct. I cannot run single codebase functions in subshells as changes of global variables in each of them would not be seen by the following one. And I want then to run the full integration test (I mean, the chain of called functions plus the test) in a subshell because the called functions are executed with `errexit` on and those might terminate the full test.

Funnily enough, I discovered this morning that `return` inside a subshell inside a function does not return the function, but only the subshell. I was not aware of that and I would have expected it to be different.

PR #78, opened 2025-06-05
Fix FIST sampler event number comparison with afterburner bug

### LINKED ISSUE #77: FIST Sampler not working in combination with Afterburner
The function [`__static__Check_If_Afterburner_Configuration_Is_Consistent_With_Sampler()`](https://github.com/smash-transport/smash-vhlle-hybrid/blob/59ed20efe05ef40b31d2752b07286632aa8ef078/bash/Afterburner_functionality.bash#L148) is currently hard-coded to compare the SMASH-hadron-sampler key `number_of_events` with the `Nevents` key from SMASH in the Afterburner stage. This leads to a crash with the following error if one uses FIST:

```
Afterburner_functionality.bash: line 155: events_sampler: unbound variable
```

I circumvented for my tests using a regex instead:
```bash
if [[ "${key}" =~ ^(number_of_events|nevents)$ ]]; then
```
If this is a reasonable approach, I can open a PR.

@AxelKrypton, the next question would be: is another hotfix for this needed? 😅

RobinSattler:
This closes #77.  As already mentioned in the issue, I implemented a new unit test for FIST checking for the inconsistency in the event number comparison between sampler and afterburner and included the check for the FIST sampler to fix the bug.

### DIFF
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -144,8 +144,18 @@ function __static__Check_If_Afterburner_Configuration_Is_Consistent_With_Sampler
     local -r config_afterburner="${HYBRID_software_configuration_file[Afterburner]}"
     if Element_In_Array_Equals_To 'Sampler' "${HYBRID_given_software_sections[@]}"; then
         local -r config_sampler="${HYBRID_software_configuration_file[Sampler]}"
+        local number_of_events_config_key_sampler
+        if [[ "${HYBRID_module[Sampler]}" = 'SMASH' ]]; then
+            number_of_events_config_key_sampler='number_of_events'
+        elif [[ "${HYBRID_module[Sampler]}" = 'FIST' ]]; then
+            number_of_events_config_key_sampler='nevents'
+        else
+            Print_Internal_And_Exit 'The used sampler module ' --emph "${HYBRID_module[Sampler]}" \
+                ' is not recognized by the function\n' --emph "${FUNCNAME}" \
+                '. This should not have happened.'
+        fi
         while read key value; do
-            if [[ "${key}" = 'number_of_events' ]]; then
+            if [[ "${key}" = "${number_of_events_config_key_sampler}" ]]; then
                 local events_sampler
                 events_sampler="${value}"
             fi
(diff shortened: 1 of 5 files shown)


bash/Afterburner_functionality.bash
```diff
@@ -144,24 +144,35 @@ function __static__Check_If_Afterburner_Configuration_Is_Consistent_With_Sampler
...
+        local number_of_events_config_key_sampler
+        if [[ "${HYBRID_module[Sampler]}" = 'SMASH' ]]; then
+            number_of_events_config_key_sampler='number_of_events'
+        elif [[ "${HYBRID_module[Sampler]}" = 'FIST' ]]; then
+            number_of_events_config_key_sampler='nevents'
+        else
+            Print_Internal_And_Exit 'The used sampler module ' --emph "${HYBRID_module[Sampler]}" \
+                ' is not recognized by the function\n' --emph "${FUNCNAME}" \
+                '\nand hence the consistency check between the number of events used in the ' \
+                'sampler and the afterburner stage cannot be performed.'
```
AxelKrypton: I believe that this error can be more about unreachable code than explanation to the user.

```suggestion
            Print_Internal_And_Exit 'The used sampler module ' --emph "${HYBRID_module[Sampler]}" \
                ' is not recognized by the function\n' --emph "${FUNCNAME}" \
                '. This should not have happened.'
```

docs/user/CHANGELOG/index.md
```diff
@@ -44,6 +44,13 @@ Given a version number `X.Y.Z`,
     This symbol indicates _changes that deserve particular attention by the user_.
 
 
+### SMASH-vHLLE-hybrid-2.1.3
+
+???+ success "&nbsp; :date: &nbsp; Release date: 2025-06-05 &emsp; :left_right_arrow: &nbsp; [Compare changes to previous version](https://github.com/smash-transport/smash-vhlle-hybrid/compare/SMASH-vHLLE-hybrid-2.1.2...SMASH-vHLLE-hybrid-2.1.3)"
+
+    :sos: &nbsp; Fix handler crash when using the FIST sampler, which was due to the consistency check of the number of events of the `Sampler` and the `Afterburner` stage.
```
AxelKrypton: The problem was only occurring when running the afterburner in combination with the FIST sampler together, isn't it? 🤔 So I would rephrase this to be more precise. Something like:

```suggestion
    :sos: &nbsp; Fix hybrid handler crash accessing an unbound variable when trying to run both the `Sampler` and the `Afterburner` stages using the FIST sampler.
```

AxelKrypton (approved):
Thanks for the fix. I would rephrase a couple of sentences, but nothing prevents me from approving already. 👍 Feel free to merge once adjusted and you can also force-push before doing so. I will then finish the hot-fix.

PR #79, opened 2025-07-14
Add new SMASH sampler config key that handles hydrodynamic simulations in Cartesian coordinates

RobinSattler:
I introduced the new config key `hydro_coordinate_system` to the allowed SMASH-hadron-sampler config keys which specifies in which coordinate system the hydrodynamic simulation was run and hence the four-position of freeze-out hypersurface elements is provided. This new config key was introduced in sampler PR smash-transport/smash-hadron-sampler#54 (possible key values: `tau-eta` (default) and `cartesian`).

As soon as the vHLLE config key is introduced to set the coordinate system of the hydro stage, it might be worth to implement a check to the handler if the keys are equal in the hydro and sampler config sections when both stages are ran (with one handler config).

@RenHirayama and @zpauliny, this concludes the implementation of the proper sampler handling for the Cartesian vHLLE runs into the hybrid framework.

### DIFF
--- a/bash/Sampler_functionality_SMASH.bash
+++ b/bash/Sampler_functionality_SMASH.bash
@@ -47,7 +47,7 @@ function Validate_Configuration_File_Of_SMASH()
         'ratio_pressure_energydensity'
     )
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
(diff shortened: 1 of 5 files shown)


AxelKrypton (approved):
The implementation looks good to me. No comments: 🟢 👍

Issue #83, opened 2025-08-04
Add python formatting and remove bare `except` statements


RobinSattler:
While playing around with a python linter/formatter, I realized that there is no unified formatting in the Hybrid-handler codebase. Additionally, there are quite a few bare `except` statements in the python scripts that should be removed, i.e. extended by the appropriate exception. Since the part of the codebase written in python is not negligibly small, it might make sense to include python in the formatting process. 

I would tackle the bare `except` removal and we should discuss the python formatter issue.

Furthermore, I also noticed that some python files miss the copyright statement. In the course of this issue, it would be good to add the statements.

## Todos

- [x] Remove bare `except` statements
- [x] Add copyright statements to python files
- [x] Add python formatter to the codebase

PR #84, opened 2025-09-01
Remove bare exceptions in python files

RobinSattler:
This tackles the first half of #83, removing the bare `except` statements from the python files. Additionally, I did some minor cleanups here and there, ending up removing some redundant code in _tests/mocks/smash_afterburner_black-box.py_. The two variables `dir_config` and `n_events_config` were unused since the corresponding `if` statement
```python
if not (dir_config and n_events_config):
```
was removed in commit 4ce1e3d (see [line 161](https://github.com/smash-transport/smash-vhlle-hybrid/commit/4ce1e3dc8b8555a47024e3c626b9d6605f9a7662#diff-92f3ef6d37bf5f68af5893006886b73f0958ecec2c1fbbc7042260831b43a9a1L161)).

### DIFF
--- a/python/add_spectators.py
+++ b/python/add_spectators.py
@@ -52,7 +52,7 @@ def get_initial_nucleons_from_config():
 
         for particle in target_particles.keys():
             N_initial_nucleons += target_particles[particle]
-    except:
+    except KeyError:
         print("The config file does not contain the necessary information "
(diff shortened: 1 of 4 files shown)


AxelKrypton (approved):
This changes seems very reasonable and improve code quality making the catches more specific. 🟢

PR #85, opened 2025-09-03
Add DOI to main documentation page

### LINKED ISSUE #81: Include DOI in main documentation page
Since we now have a DOI associated to the project, it makes sense to enhance visibility in the first page of the documentation. This issue can be used to accomplish a few related tasks:

- [x] Change title to `Hybrid-handler` both in documentation and in the README file.
- [x] Grep for missing dash between `Hybrid` and `handler`.
- [x] Use generic names for stages and use `vHLLE` as example for `Hydro`.
  * List all supported software with name only in title page
- [x] Update the admonition box about citing the code adding the information about DOI. It is probably a good idea to have there an example for how to cite the handler (DOI from handler + 4 citations for each stage).
- [x] Change `SMASH Hybrid Team` in copyright statements to `Hybrid-handler Team`

RobinSattler:
This closes #81. @AxelKrypton, since these are quite a lot of changes, it might be easier to go through the commits one by one. For this purpose, I grouped similar changes into a commit, e.g., all the changes from `hybrid handler` to `Hybrid-handler` are in one commit and nothing else is changed.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -8,19 +8,33 @@ hide:
 
 ![Image title](images/logo.png){ width="25%", align=right }
 
-Event-by-event hybrid model for the description of relativistic heavy-ion collisions in the low and high baryon-density regime. This model constitutes a chain of different submodules to appropriately describe each phase of the collision with its corresponding degrees of freedom. It consists of the following modules:
+Event-by-event hybrid model for the description of relativistic heavy-ion collisions in the low and high baryon-density regime.
+This model constitutes a chain of different submodules to appropriately describe each phase of the collision with its corresponding degrees of freedom.
+It consists of the following stages and supported software modules are listed within each stage:
 
-:cloud_tornado: &nbsp; **SMASH** hadronic transport approach to provide the initial conditions
+:cloud_tornado: &nbsp; **Initial conditions**
 
-:droplet: &nbsp; **vHLLE** 3+1D viscous hydrodynamics approach to describe the evolution of the hot and dense fireball
+:   :simple-ticktick: &nbsp; *SMASH* (hadronic transport approach)
 
-:   :material-arrow-right-bottom: &nbsp; CORNELIUS tool to construct a hypersurface of constant energy density from the hydrodynamical evolution (embedded in vHLLE)
+:droplet: &nbsp; **Hydrodynamics** to describe the evolution of the hot and dense fireball
 
-:seedling: &nbsp; **Sampler** to perform Cooper-Frye particlization of the elements on the freezeout hypersurface
+:   :simple-ticktick: &nbsp; *vHLLE* (3+1D viscous hydrodynamics approach)
+    :   :material-arrow-right-bottom: &nbsp; including *CORNELIUS* to construct a constant energy density hypersurface from the hydrodynamical evolution
 
-:fire: &nbsp; **SMASH** hadronic transport approach to perform the afterburner evolution
+:seedling: &nbsp; **Hadron sampler** to perform Cooper-Frye particlization of the elements on the freezeout hypersurface
 
-!!! info "Give credit appropriately"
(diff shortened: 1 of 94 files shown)


docs/user/CHANGELOG/index.md
```diff
@@ -99,4 +99,4 @@ Given a version number `X.Y.Z`,
 
 ### SMASH-vHLLE-hybrid-1.0
 
-!!! success ":date: &nbsp; Release date: 2020-11-18 &emsp; :left_right_arrow: &nbsp; [First public version of the SMASH-vHLLE-hybrid](https://github.com/smash-transport/smash-vhlle-hybrid/releases/tag/SMASH-vHLLE-hybrid-1.0)"
+!!! success ":date: &nbsp; Release date: 2020-11-18 &emsp; :left_right_arrow: &nbsp; [First public version of the Hybrid-handler](https://github.com/smash-transport/smash-vhlle-hybrid/releases/tag/SMASH-vHLLE-hybrid-1.0)"
```
RenHirayama: I think this should remain with the old name, I find it weird to change the CHANGELOG after the fact.

docs/index.md
```diff
@@ -22,5 +22,5 @@ Event-by-event hybrid model for the description of relativistic heavy-ion collis
 
 !!! info "Give credit appropriately"
 
-    If you are using the SMASH-vHLLE-hybrid, please cite [Eur.Phys.J.A 58(2022)11,230](https://arxiv.org/abs/2112.08724).
+    If you are using the Hybrid-handler, please cite [Eur.Phys.J.A 58(2022)11,230](https://arxiv.org/abs/2112.08724).
```
RenHirayama: I think this should remain with the old name

RenHirayama (commented):
I don't agree with changing the team name but no strong feelings

docs/user/CHANGELOG/index.md
AxelKrypton: I agree. This was Anna's work. We should possibly clarify in the CHANGELOG itself within this PR that the historical name is not any more used in favour of the new more generic `Hybrid-handler` one.

docs/index.md
```diff
@@ -4,23 +4,37 @@ hide:
...
+:fire: &nbsp; **Afterburner** evolution
+
+:   :simple-ticktick: &nbsp; *SMASH* (hadronic transport approach)
+
+!!! info "Give credit appropriately &nbsp; [![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.15880337.svg){ align=right }](https://doi.org/10.5281/zenodo.15880337)"
+
+    If you are using the Hybrid-handler, please cite the corresponding [:simple-doi: software DOI](https://doi.org/10.5281/zenodo.15880337) of the specific code version employed.
+    Depending on the stages and modules you are using, please cite the appropriate papers and software DOIs as well.
+
+    For example, if you use the SMASH-vHLLE-Hybrid, besides citing the Hybrid-handler itself, you should cite [:newspaper: Eur.Phys.J.A 58(2022)11, 230](https://arxiv.org/abs/2112.08724) for the general approach.
```
AxelKrypton: I would rather change this to something like that makes it clear what `SMASH-vHLLE-Hybrid` means. 🤔 I would then remove _about the SMASH-vHLLE-Hybrid approach_ in the following sentence.



```suggestion
    For example, if you use the SMASH-vHLLE-Hybrid approach, besides citing the Hybrid-handler itself, you should cite [:newspaper: Eur.Phys.J.A 58(2022)11, 230](https://arxiv.org/abs/2112.08724) for physical aspects.
```

AxelKrypton (approved):
I believe that renaming the team makes sense. I have a suggestion about the index page and one about the CHANGELOG. @RobinSattler I already approve. You can ping on me about this on Mattermost if you need a further feedback.

PR #86, opened 2025-11-16
Add_corona functionality for dynflu

RenHirayama:
This includes the `add_corona.py` script necessary for DynFlu, including example configs.

### DIFF
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -15,7 +15,7 @@ function Prepare_Software_Input_File_Afterburner()
     Ensure_Given_Files_Exist "${HYBRID_software_base_config_file[Afterburner]}"
     Copy_Base_Configuration_To_Output_Folder_For 'Afterburner'
     Replace_Keys_In_Configuration_File_If_Needed_For 'Afterburner'
-    __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Without_Spectators
+    __static__Create_Sampled_Particles_File_Or_Symbolic_Link
     __static__Check_If_Afterburner_Configuration_Is_Consistent_With_Sampler
 }
 
@@ -90,7 +90,30 @@ function __static__Run_Add_Spectators_Python_Script()
         '--smash_config' "${HYBRID_software_output_directory[IC]}/config.yaml" 2> /dev/null
 }
 
-function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Without_Spectators()
+function __static__Run_Add_Corona_Python_Script()
+{
+    local python_fatal_value_error
+    # NOTE: To extract the right error code in the python script, it has to be handed to it right
+    #       before calling the script since the error codes are not exported by default.
+    python_fatal_value_error=${HYBRID_fatal_value_error} \
+        "${HYBRID_external_python_scripts[Add_corona_from_IC_and_Hydro]}" \
+        '--sampled_particle_list' "${HYBRID_software_input_file[Afterburner]}" \
+        '--corona_particle_lists' "${HYBRID_software_input_file[IC_corona]}" "${HYBRID_software_input_file[Hydro_corona]}" \
+        '--output_file' "${target_link_name}" 2> /dev/null
+}
+
+function __static__Create_Symbolic_Link()
+{
+    if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
+        ln -s -f "${HYBRID_software_input_file[Afterburner]}" "${target_link_name}"
+    elif [[ ! "${target_link_name}" -ef "${HYBRID_software_input_file[Afterburner]}" ]]; then
+        exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
+            'File ' --emph "${target_link_name}" ' exists but it is not the Afterburner input file ' \
+            --emph "${HYBRID_software_input_file[Afterburner]}" ' to be used.'
+    fi
+}
+
+function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
 {
     local -r target_link_name="${HYBRID_software_output_directory[Afterburner]}/${HYBRID_afterburner_list_filename}"
     if [[ "${HYBRID_optional_feature[Add_spectators_from_IC]}" = 'TRUE' ]]; then
@@ -128,14 +151,35 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Withou
                     'Adding spectators from the IC particles to the sampled particles file failed.'
             fi
         fi
-    else
-        if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
-            ln -s -f "${HYBRID_software_input_file[Afterburner]}" "${target_link_name}"
-        elif [[ ! "${target_link_name}" -ef "${HYBRID_software_input_file[Afterburner]}" ]]; then
-            exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
-                'File ' --emph "${target_link_name}" ' exists but it is not the Afterburner input file ' \
-                --emph "${HYBRID_software_input_file[Afterburner]}" ' to be used.'
+    elif [[ "${HYBRID_optional_feature[Add_corona_from_IC_and_Hydro]}" = 'TRUE' ]]; then
+        Ensure_Given_Files_Do_Not_Exist "${target_link_name}"
+        Ensure_Given_Files_Exist \
+            "${HYBRID_software_input_file[Afterburner]}" \
+            "${HYBRID_software_input_file[IC_corona]}"
+        local python_exit_code
+        if __static__Run_Add_Corona_Python_Script; then
+            # BE AWARE: The negation of the if-clause does not work because then the exit code cannot
+            #           be extracted properly (it will always be 0 because of the true if-statement).
+            :
+        else
+            python_exit_code=$?
+            if [[ ${python_exit_code} -eq ${HYBRID_fatal_value_error} ]]; then
+                exit_code=${HYBRID_fatal_value_error} Print_Fatal_And_Exit \
+                    'Something is wrong with ' \
+                    --emph "Add_corona_from_IC_and_Hydro: true" '.'
+            elif [[ ${python_exit_code} -eq 2 ]]; then
+                Print_Internal_And_Exit \
+                    'The handing over of the ' --emph "python_fatal_value_error" \
+                    ' to the Python script that adds corona from the IC and Hydro did not work.'
+            elif [[ ${python_exit_code} -eq 9 ]]; then
+                __static__Create_Symbolic_Link
+            else
+                Print_Fatal_And_Exit \
+                    'Adding corona from IC and Hydro to the sampled particles file failed.'
+            fi
         fi
+    else
+        __static_Create_Symbolic_Link
     fi
 }
 
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -179,7 +179,13 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
             fi
         fi
     else
-        __static_Create_Symbolic_Link
+        if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
+            ln -s -f "${HYBRID_software_input_file[Afterburner]}" "${target_link_name}"
+        elif [[ ! "${target_link_name}" -ef "${HYBRID_software_input_file[Afterburner]}" ]]; then
+            exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
+                'File ' --emph "${target_link_name}" ' exists but it is not the Afterburner input file ' \
+                --emph "${HYBRID_software_input_file[Afterburner]}" ' to be used.'
+        fi
     fi
 }
 
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -92,14 +92,12 @@ function __static__Run_Add_Spectators_Python_Script()
 
 function __static__Run_Add_Corona_Python_Script()
 {
-    local python_fatal_value_error
-    # NOTE: To extract the right error code in the python script, it has to be handed to it right
-    #       before calling the script since the error codes are not exported by default.
-    python_fatal_value_error=${HYBRID_fatal_value_error} \
-        "${HYBRID_external_python_scripts[Add_corona_from_IC_and_Hydro]}" \
+    "${HYBRID_external_python_scripts[Add_corona_from_IC_and_Hydro]}" \
         '--sampled_particle_list' "${HYBRID_software_input_file[Afterburner]}" \
-        '--corona_particle_lists' "${HYBRID_software_input_file[IC_corona]}" "${HYBRID_software_input_file[Hydro_corona]}" \
+        '--corona_particle_lists' "${HYBRID_software_input_file[IC_corona]}" \
+        "${HYBRID_software_input_file[Hydro_corona]}" \
         '--output_file' "${target_link_name}" 2> /dev/null
+    Print_Info $?
 }
 
 function __static__Create_Symbolic_Link()
@@ -158,20 +156,17 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
             "${HYBRID_software_input_file[IC_corona]}"
         local python_exit_code
         if __static__Run_Add_Corona_Python_Script; then
-            # BE AWARE: The negation of the if-clause does not work because then the exit code cannot
-            #           be extracted properly (it will always be 0 because of the true if-statement).
             :
         else
             python_exit_code=$?
-            if [[ ${python_exit_code} -eq ${HYBRID_fatal_value_error} ]]; then
-                exit_code=${HYBRID_fatal_value_error} Print_Fatal_And_Exit \
-                    'Something is wrong with ' \
-                    --emph "Add_corona_from_IC_and_Hydro: true" '.'
-            elif [[ ${python_exit_code} -eq 2 ]]; then
+            Print_Info \
+                'Exit code:' "${python_exit_code}"
+            if [[ ${python_exit_code} -eq 3 ]]; then
                 Print_Internal_And_Exit \
-                    'The handing over of the ' --emph "python_fatal_value_error" \
-                    ' to the Python script that adds corona from the IC and Hydro did not work.'
-            elif [[ ${python_exit_code} -eq 9 ]]; then
+                    'Internal error:' "${target_link_name}" 'found to exist even if it was ensured not to.'
+            elif [[ ${python_exit_code} -eq 4 ]]; then
+                Print_Warning \
+                    'No corona particles were found, making a symbolic link for the sampled particles.'
                 __static__Create_Symbolic_Link
             else
                 Print_Fatal_And_Exit \
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -113,7 +113,37 @@ function __static__Create_Symbolic_Link()
 function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
 {
     local -r target_link_name="${HYBRID_software_output_directory[Afterburner]}/${HYBRID_afterburner_list_filename}"
-    if [[ "${HYBRID_optional_feature[Add_spectators_from_IC]}" = 'TRUE' ]]; then
+    if [[ "${HYBRID_optional_feature[Add_corona_from_IC_and_Hydro]}" = 'TRUE' ]]; then
+        Ensure_Given_Files_Do_Not_Exist "${target_link_name}"
+        Ensure_Given_Files_Exist \
+            "${HYBRID_software_input_file[Afterburner]}" \
+            "${HYBRID_software_input_file[IC_corona]}"
+        local python_exit_code
+        if __static__Run_Add_Corona_Python_Script; then
+            :
+        else
+            python_exit_code=$?
+            if [[ ${python_exit_code} -eq 2 ]]; then
+                Print_Internal_And_Exit \
+                    --emph "${HYBRID_software_input_file[IC_corona]}" \
+                    'or' --emph "${HYBRID_software_input_file[Afterburner]}" \
+                    'not found even if it was ensured to exist.'
+            elif [[ ${python_exit_code} -eq 3 ]]; then
+                Print_Fatal_And_Exit \
+                    'Could not determine number of events from last line of at least one input file.'
+            elif [[ ${python_exit_code} -eq 4 ]]; then
+                Print_Warning \
+                    'No corona particles were found, creating a symbolic link for the sampled particles.'
+                __static__Create_Symbolic_Link
+            elif [[ ${python_exit_code} -eq 5 ]]; then
+                Print_Internal_And_Exit \
+                    --emph "${target_link_name}" 'already exists even if it was ensured not to.'
+            else
+                Print_Fatal_And_Exit \
+                    'Adding corona from IC and Hydro to the sampled particles file failed.'
+            fi
+        fi
+    elif [[ "${HYBRID_optional_feature[Add_spectators_from_IC]}" = 'TRUE' ]]; then
         Ensure_Given_Files_Do_Not_Exist "${target_link_name}"
         # Here the config.yaml file is expected to be produced by SMASH in the output folder
         # during the IC run. It is used to determine the initial number of particles.
@@ -146,36 +176,6 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
             else
                 Print_Fatal_And_Exit \
                     'Adding spectators from the IC particles to the sampled particles file failed.'
-            fi
-        fi
-    elif [[ "${HYBRID_optional_feature[Add_corona_from_IC_and_Hydro]}" = 'TRUE' ]]; then
-        Ensure_Given_Files_Do_Not_Exist "${target_link_name}"
-        Ensure_Given_Files_Exist \
-            "${HYBRID_software_input_file[Afterburner]}" \
-            "${HYBRID_software_input_file[IC_corona]}"
-        local python_exit_code
-        if __static__Run_Add_Corona_Python_Script; then
-            :
-        else
-            python_exit_code=$?
-            if [[ ${python_exit_code} -eq 2 ]]; then
-                Print_Internal_And_Exit \
-                    --emph "${HYBRID_software_input_file[IC_corona]}" \
-                    'or' --emph "${HYBRID_software_input_file[Afterburner]}" \
-                    'not found even if it was ensured to exist.'
-            elif [[ ${python_exit_code} -eq 3 ]]; then
-                Print_Fatal_And_Exit \
-                    'Could not determine number of events from last line of some input file.'
-            elif [[ ${python_exit_code} -eq 4 ]]; then
-                Print_Warning \
-                    'No corona particles were found, making a symbolic link for the sampled particles.'
-                __static__Create_Symbolic_Link
-            elif [[ ${python_exit_code} -eq 5 ]]; then
-                Print_Internal_And_Exit \
-                    --emph "${target_link_name}" 'already exists even if it was ensured not to.'
-            else
-                Print_Fatal_And_Exit \
-                    'Adding corona from IC and Hydro to the sampled particles file failed.'
             fi
         fi
     else
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -97,7 +97,6 @@ function __static__Run_Add_Corona_Python_Script()
         '--corona_particle_lists' "${HYBRID_software_input_file[IC_corona]}" \
         "${HYBRID_software_input_file[Hydro_corona]}" \
         '--output_file' "${target_link_name}" 2> /dev/null
-    Print_Info $?
 }
 
 function __static__Create_Symbolic_Link()
@@ -159,28 +158,28 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
             :
         else
             python_exit_code=$?
-            Print_Info \
-                'Exit code:' "${python_exit_code}"
-            if [[ ${python_exit_code} -eq 3 ]]; then
+            if [[ ${python_exit_code} -eq 2 ]]; then
                 Print_Internal_And_Exit \
-                    'Internal error:' "${target_link_name}" 'found to exist even if it was ensured not to.'
+                    --emph "${HYBRID_software_input_file[IC_corona]}" \
+                    'or' --emph "${HYBRID_software_input_file[Afterburner]}" \
+                    'not found even if it was ensured to exist.'
+            elif [[ ${python_exit_code} -eq 3 ]]; then
+                Print_Fatal_And_Exit \
+                    'Could not determine number of events from last line of some input file.'
             elif [[ ${python_exit_code} -eq 4 ]]; then
                 Print_Warning \
                     'No corona particles were found, making a symbolic link for the sampled particles.'
                 __static__Create_Symbolic_Link
+            elif [[ ${python_exit_code} -eq 5 ]]; then
+                Print_Internal_And_Exit \
+                    --emph "${target_link_name}" 'already exists even if it was ensured not to.'
             else
                 Print_Fatal_And_Exit \
                     'Adding corona from IC and Hydro to the sampled particles file failed.'
             fi
         fi
     else
-        if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
-            ln -s -f "${HYBRID_software_input_file[Afterburner]}" "${target_link_name}"
-        elif [[ ! "${target_link_name}" -ef "${HYBRID_software_input_file[Afterburner]}" ]]; then
-            exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
-                'File ' --emph "${target_link_name}" ' exists but it is not the Afterburner input file ' \
-                --emph "${HYBRID_software_input_file[Afterburner]}" ' to be used.'
-        fi
+        __static__Create_Symbolic_Link
     fi
 }
 
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -118,30 +118,38 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
         Ensure_Given_Files_Exist \
             "${HYBRID_software_input_file[Afterburner]}" \
             "${HYBRID_software_input_file[IC_corona]}"
-        local python_exit_code
+        # NOTE: Since the errexit option enabled, the python script to add the spectators is run in the if-statement
+        #       and the possible exit code is accessed at the very beginning of the else-clause.
         if __static__Run_Add_Corona_Python_Script; then
+            # BE AWARE: The negation of the if-clause does not work because then the exit code cannot
+            #           be extracted properly (it will always be 0 because of the true if-statement).
             :
         else
-            python_exit_code=$?
-            if [[ ${python_exit_code} -eq 2 ]]; then
-                Print_Internal_And_Exit \
-                    --emph "${HYBRID_software_input_file[IC_corona]}" \
-                    'or' --emph "${HYBRID_software_input_file[Afterburner]}" \
-                    'not found even if it was ensured to exist.'
-            elif [[ ${python_exit_code} -eq 3 ]]; then
-                Print_Fatal_And_Exit \
-                    'Could not determine number of events from last line of at least one input file.'
-            elif [[ ${python_exit_code} -eq 4 ]]; then
-                Print_Warning \
-                    'No corona particles were found, creating a symbolic link for the sampled particles.'
-                __static__Create_Symbolic_Link
-            elif [[ ${python_exit_code} -eq 5 ]]; then
-                Print_Internal_And_Exit \
-                    --emph "${target_link_name}" 'already exists even if it was ensured not to.'
-            else
-                Print_Fatal_And_Exit \
-                    'Adding corona from IC and Hydro to the sampled particles file failed.'
-            fi
+            case $? in
+                2)
+                    Print_Internal_And_Exit \
+                        --emph "${HYBRID_software_input_file[IC_corona]}" \
+                        'or' --emph "${HYBRID_software_input_file[Afterburner]}" \
+                        'not found even if it was ensured to exist.'
+                    ;;
+                3)
+                    Print_Fatal_And_Exit \
+                        'Could not determine number of events from last line of at least one input file.'
+                    ;;
+                4)
+                    Print_Warning \
+                        'No corona particles were found, creating a symbolic link for the sampled particles.'
+                    __static__Create_Symbolic_Link
+                    ;;
+                5)
+                    Print_Internal_And_Exit \
+                        --emph "${target_link_name}" 'already exists even if it was ensured not to.'
+                    ;;
+                *)
+                    Print_Fatal_And_Exit \
+                        'Adding corona from IC and Hydro to the sampled particles file failed.'
+                    ;;
+            esac
         fi
     elif [[ "${HYBRID_optional_feature[Add_spectators_from_IC]}" = 'TRUE' ]]; then
         Ensure_Given_Files_Do_Not_Exist "${target_link_name}"
@@ -157,26 +165,28 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
         # Run Python script to add spectators
         # NOTE: Since the errexit option enabled, the python script to add the spectators is run in the if-statement
         #       and the possible exit code is accessed at the very beginning of the else-clause.
-        local python_exit_code
         if __static__Run_Add_Spectators_Python_Script; then
             # BE AWARE: The negation of the if-clause does not work because then the exit code cannot
             #           be extracted properly (it will always be 0 because of the true if-statement).
             :
         else
-            python_exit_code=$?
-            if [[ ${python_exit_code} -eq ${HYBRID_fatal_value_error} ]]; then
-                exit_code=${HYBRID_fatal_value_error} Print_Fatal_And_Exit \
-                    'It was attempted to add spectators from multiple IC events to the sampled particles file. Only' \
-                    'running one IC event is supported when using the Afterburner config key ' \
-                    --emph "Add_spectators_from_IC: true" '.'
-            elif [[ ${python_exit_code} -eq 2 ]]; then
-                Print_Internal_And_Exit \
-                    'The handing over of the ' --emph "python_fatal_value_error" \
-                    ' to the Python script that adds spectators from the IC did not work.'
-            else
-                Print_Fatal_And_Exit \
-                    'Adding spectators from the IC particles to the sampled particles file failed.'
-            fi
+            case $? in
+                ${HYBRID_fatal_value_error})
+                    exit_code=${HYBRID_fatal_value_error} Print_Fatal_And_Exit \
+                        'It was attempted to add spectators from multiple IC events to the sampled particles file. ' \
+                        'Only running one IC event is supported when using the Afterburner config key ' \
+                        --emph "Add_spectators_from_IC: true" '.'
+                    ;;
+                2)
+                    Print_Internal_And_Exit \
+                        'The handing over of the ' --emph "python_fatal_value_error" \
+                        ' to the Python script that adds spectators from the IC did not work.'
+                    ;;
+                *)
+                    Print_Fatal_And_Exit \
+                        'Adding spectators from the IC particles to the sampled particles file failed.'
+                    ;;
+            esac
         fi
     else
         __static__Create_Symbolic_Link
--- a/bash/Sampler_functionality_SMASH.bash
+++ b/bash/Sampler_functionality_SMASH.bash
@@ -45,8 +45,6 @@ function Validate_Configuration_File_Of_SMASH()
         'ecrit'
         'cs2'
         'ratio_pressure_energydensity'
-        'hydro_coordinate_system'
-        'transversal_smearing'
     )
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
         allowed_keys+=('create_root_output' 'hydro_coordinate_system')
@@ -90,7 +88,7 @@ function Validate_Configuration_File_Of_SMASH()
                 fi
                 ((keys_to_be_found--))
                 ;;
-            shear | bulk | create_root_output | transversal_smearing)
+            shear | bulk | create_root_output)
                 if [[ ! "${value}" =~ ^[01]$ ]]; then
                     Print_Error 'Key ' --emph "${key}" ' must be either ' \
                         --emph '0' ' or ' --emph '1' '.'
--- a/bash/Sampler_functionality_SMASH.bash
+++ b/bash/Sampler_functionality_SMASH.bash
@@ -45,6 +45,8 @@ function Validate_Configuration_File_Of_SMASH()
         'ecrit'
         'cs2'
         'ratio_pressure_energydensity'
+        'hydro_coordinate_system'
+        'transversal_smearing'
     )
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
         allowed_keys+=('create_root_output' 'hydro_coordinate_system')
@@ -88,10 +90,17 @@ function Validate_Configuration_File_Of_SMASH()
                 fi
                 ((keys_to_be_found--))
                 ;;
-            shear | bulk | create_root_output)
+            shear | bulk | create_root_output | transversal_smearing)
                 if [[ ! "${value}" =~ ^[01]$ ]]; then
                     Print_Error 'Key ' --emph "${key}" ' must be either ' \
                         --emph '0' ' or ' --emph '1' '.'
+                    return 1
+                fi
+                ;;
+            hydro_coordinate_system)
+                if [[ ! "${value,,}" =~ ^("tau-eta"|"milne"|"cartesian")$ ]]; then
+                    Print_Error 'Key ' --emph "${key}" ' must be either ' \
+                        --emph 'Milne' ' or ' --emph 'Cartesian' '.'
                     return 1
                 fi
                 ;;
--- a/bash/Sampler_functionality_SMASH.bash
+++ b/bash/Sampler_functionality_SMASH.bash
@@ -95,13 +95,6 @@ function Validate_Configuration_File_Of_SMASH()
                     return 1
                 fi
                 ;;
-            hydro_coordinate_system)
-                if [[ ! "${value,,}" =~ ^("tau-eta"|"milne"|"cartesian")$ ]]; then
-                    Print_Error 'Key ' --emph "${key}" ' must be either ' \
-                        --emph 'Milne' ' or ' --emph 'Cartesian' '.'
-                    return 1
-                fi
-                ;;
             number_of_events)
                 if [[ ! "${value}" =~ ^[1-9][0-9]*$ ]]; then
                     Print_Error 'Found non-integer value ' --emph "${value:-''}" \
--- a/bash/sanity_checks.bash
+++ b/bash/sanity_checks.bash
@@ -142,7 +142,7 @@ function __static__Set_Base_Configuration_File_If_Unset()
             else
                 ic_key='IC_ge_3.2'
             fi
-#            HYBRID_software_base_config_file[IC]="${HYBRID_software_base_config_file[${ic_key}]}"
+            HYBRID_software_base_config_file[IC]="${HYBRID_software_base_config_file[${ic_key}]}"
             ;;
         Sampler)
             local sampler_key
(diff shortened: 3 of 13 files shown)


bash/Sampler_functionality_SMASH.bash
```diff
@@ -95,6 +95,13 @@ function Validate_Configuration_File_Of_SMASH()
                     return 1
                 fi
                 ;;
+            hydro_coordinate_system)
+                if [[ ! "${value,,}" =~ ^("tau-eta"|"milne"|"cartesian")$ ]]; then
+                    Print_Error 'Key ' --emph "${key}" ' must be either ' \
+                        --emph 'Milne' ' or ' --emph 'Cartesian' '.'
+                    return 1
+                fi
+                ;;
```
RobinSattler: This should not be in here. It's already implemented (see lines 112 to 119 down below).

docs/user/CHANGELOG/index.md
```diff
@@ -48,11 +48,13 @@ Given a version number `X.Y.Z`,
...
     **Changes:**
 
+    :new: &nbsp; Added `Add_corona_from_IC_and_Hydro` configuration key in the `Afterburner` module, which merges the files containing corona particles from :file_folder: ***IC*** and :file_folder: ***Hydro*** into the particle lists sampled in :file_folder: ***Sampler***. For that, the `Particles` output with `Format: [Oscar2013]` must be active in the `IC` configuration. 
+
     :new: &nbsp; Added new valid config key `hydro_coordinate_system` for SMASH sampler to handle hydrodynamic simulations in Cartesian coordinates properly, additional to the runs in tau-eta frame.
 
     :new: &nbsp; The `Input_file` configuration key in the `Hydro` and `Afterburner` modules can now be used to modify the expected file names from the appropriate :file_folder: ***IC*** and :file_folder: ***Sampler*** sub-folders, when a string without a `/` character is given.
 
-
+
```
RobinSattler: Remove white space.

configs/vhlle_hydro_cartesian
RobinSattler, line 1: Wouldn't the file name `vhlle_hydro_DynFlu` or similar be more accurate? 🤔

RobinSattler:
I wonder if we should wait with the merge of this PR until the dynamical fluidization is implemented in vHLLE `main` branch. If I'm not mistaken it's planned to have a config key in vHLLE to choose between hyperbolic and dynamical fluidization runs, right? In this case, we would have to alter the configs you included when the vHLLE work is done if we finish this PR before.

configs/hadron_sampler_cartesian
RobinSattler, line 1: I'd be in favor of naming this file `hadron_sampler_DynFlu` since it's more consistent and hints at the usage of the file. And is there a reason why you are not listing all keys? Since there are not many, this would make it easier for the user to copy it somewhere and use it as a base config.

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+def read_corona_file(file_path):
+    with open(file_path, 'r') as f:
+        lines = [line.strip().split() for line in f if not line.startswith('#')]
+    array = np.atleast_2d(np.array(lines, dtype=str)) if len(lines) != 0 else []
+    return array
+
+def initialize_final_file(file_path):
+    header_string = "#!OSCAR2013 particle_lists t x y z mass p0 px py pz pdg ID charge \n\
+# Units: fm fm fm fm GeV GeV GeV GeV GeV none none e \n\
+# SMASH-3.3 \n"
```
RobinSattler: I'm not happy with hard-coding a SMASH version here. Maybe it's smarter to use the same procedure as in `add_spectators.py`, in which the entire _particle_lists.oscar_ output file from the sampler is read and written to a new file and the spectators are just attached to the end of the file. This is also not the sweetest solution, e.g., when runs are continued with old lists, but better than a hard-coded version. And this might be more effort if there is more than one event.

RenHirayama:
Up to you, either way it should be simple from the handler side. We will have a dynflu meeting this Tuesday, I will let you know.

docs/user/CHANGELOG/index.md
```diff
@@ -48,11 +48,13 @@ Given a version number `X.Y.Z`,
 
     **Changes:**
 
+    :new: &nbsp; Added `Add_corona_from_IC_and_Hydro` configuration key in the `Afterburner` module, which merges the files containing corona particles from :file_folder: ***IC*** and :file_folder: ***Hydro*** into the particle lists sampled in :file_folder: ***Sampler***. For that, the `Particles` output with `Format: [Oscar2013]` must be active in the `IC` configuration.
```
RobinSattler: The last sentence should not be in the changelog, but in the documentation.

configs/predef_configs/config_AuAu_4.3_DynFlu.yaml
```diff
@@ -0,0 +1,33 @@
...
+            Collider:
+                Projectile:
+                    Particles: {2212: 79, 2112: 118} #Gold197
+                Target:
+                    Particles: {2212: 79, 2112: 118} #Gold197
+                Sqrtsnn: 4.3
+
+Hydro:
+    Executable: /path/to/vHLLE
+    Config_file: /path/to/handler/configs/vhlle_hydro_cartesian
```
RobinSattler: Shouldn't this be the default from the _configs_ folder then?

configs/predef_configs/config_AuAu_4.3_DynFlu.yaml
```diff
@@ -0,0 +1,33 @@
...
+    Input_file: SMASH_IC.oscar
+    Software_keys:
+        Gaussian_Sigma: 0.6
+        minParticlesFO: 15
+        nx: 50
+        ny: 50
+        nz: 50
```
RobinSattler: Is this fine enough? 🤔 The cells are quite large then, aren't they?

configs/smash_initial_conditions_DynFlu.yaml
```diff
@@ -0,0 +1,44 @@
...
+        Projectile:
+            Particles: {2212: 79, 2112: 118} #Gold197
+        Target:
+            Particles: {2212: 79, 2112: 118} #Gold197
+        Sqrtsnn: 7.7
+
+        Fermi_Motion: frozen
+
+        Impact:
+            Range: [0.0,3.3]
```
RobinSattler: I think, I'd put the impact key in the example handler config _configs/predef_configs/config_AuAu_4.3_DynFlu.yaml_ as well.

configs/smash_initial_conditions_DynFlu.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+        Initial_Conditions:
+            Type: "Dynamic"
+            Energy_Density_Threshold: 0.5
+            Formation_Time_Fraction: 0.2
+            Fluidizable_Processes: ["All"]
+            Delay_Initial_Elastic: true
```
RobinSattler, line 38: Maybe some/all of them too.

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+    with open(file_path, 'w') as f:
+        f.write(header_string)
+
+def read_Sampler_file(file_path, file_path_new, array_to_append):
```
RobinSattler: I'm not sure if we have a convention for it, but I'd be in favor of all lower-case function name since all the others are.

```suggestion
def read_sampler_file(file_path, file_path_new, array_to_append):
```

python/add_corona.py
RobinSattler, line 1: There is white space in this file which could be removed. And more importantly, how is this file dealing with multiple IC events? It looks to me like this would lead to a messed up merge of the IC corona file into the sampled particle list.

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+    args = parser.parse_args()
+
+    output = args.output_file
+    if os.path.isfile(output):
+        print(output+" already exists!", file=sys.stderr)
+        if not args.force:
+            sys.exit(3)
+
+    is_any_corona_file_present = False
+    corona = []
```
RobinSattler: I'd like it more verbose, e.g.,

```suggestion
    corona_particles = []
```

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+            print(corona_file+" not present!", file=sys.stderr)
+            continue
+        is_any_corona_file_present = True
+        if len(corona) == 0:
+            file_read = read_corona_file(corona_file)
+            if len(file_read) != 0:
+               corona = file_read[:,:12]
+        else:
+            corona = np.r_[corona,read_corona_file(corona_file)[:,:12]]
+    if len(corona) == 0:
```
RobinSattler: I think, you should also check if no corona file was found and give a separate error.
```suggestion
    if not is_any_corona_file_present:
        # error message about no corona files found
        sys.exit()
    elif len(corona) == 0:
```

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+
+    output = args.output_file
+    if os.path.isfile(output):
+        print(output+" already exists!", file=sys.stderr)
+        if not args.force:
+            sys.exit(3)
+
+    is_any_corona_file_present = False
+    corona = []
+    for i_file,corona_file in enumerate(args.corona_particle_lists):
```
RobinSattler: Maybe I'm overlooking something, but why use `enumerate` if you never use the index `i_file`? 🤔

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+            continue
+        is_any_corona_file_present = True
+        if len(corona) == 0:
+            file_read = read_corona_file(corona_file)
+            if len(file_read) != 0:
+               corona = file_read[:,:12]
+        else:
+            corona = np.r_[corona,read_corona_file(corona_file)[:,:12]]
+    if len(corona) == 0:
+        print("no corona particles!", file=sys.stderr)
```
RobinSattler:
```suggestion
        print("No corona particles!", file=sys.stderr)
```

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+
+def read_Sampler_file(file_path, file_path_new, array_to_append):
+    # Open the file and read lines
+    lines_buffer = np.empty((0, 12), dtype=str)
+    event_no = 0
+    with open(file_path, 'r') as file:
+        for line in file:
+            if '#!OSCAR2013' in line or 'Units:' in line or 'SMASH-' in line:
+                continue
+            if 'end' in line:
```
RobinSattler: Is there a reason why you didn't choose (the safer) `elif` here?
```suggestion
            elif 'end' in line:
```

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+                # Convert the accumulated lines to a NumPy array when 'end' is encountered
+                one_event_array = lines_buffer
+                #print(f"NumPy Array up until 'end': {one_event_array}")
```
RobinSattler: These two lines seem unnecessary since `array = lines_buffer` is set in line 50 (and `one_event_array` is never used).

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+            
+                # Join IC and Sampler
+                first_event_line = "# event {} out \n".format(event_no)
+                final_event_line = "# event {} end \n".format(event_no)
+                
+                with open(file_path_new, 'a') as f:
+                    f.write(first_event_line)
+                    
+                array = lines_buffer
```
RobinSattler: I think, I'd prefer `one_event_array`.
```suggestion
                one_event_array = lines_buffer
```

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+                with open(file_path_new, 'a') as f:
+                    np.savetxt(f, array, delimiter=' ', fmt='%s')
+                with open(file_path_new, 'a') as f:
+                    np.savetxt(f, array_to_append, delimiter=' ', fmt='%s')
```
RobinSattler: Doesn't
```python
np.savetxt(f, array_to_append, delimiter=' ', fmt='%s')
```
dump the whole corona particles (of all events) in each event?

python/add_corona.py
```diff
@@ -0,0 +1,99 @@
...
+                    f.write(final_event_line)
+                # Reset the buffer for the next set of lines
+                lines_buffer = np.empty((0, 12), dtype=str)
+                # Increase event number
+                event_no += 1
+
+            else:
+                if not line.strip().startswith('#'):
+                    line_array = np.array(line.split())
+                    lines_buffer = np.vstack([lines_buffer, line_array])
```
RobinSattler: For the sake of readability, I'd prefer this to be the first `if` and the second check the `elif 'end' in line`.

bash/Afterburner_functionality.bash
```diff
@@ -128,14 +148,30 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Withou
...
+            "${HYBRID_software_input_file[Afterburner]}" \
+            "${HYBRID_software_input_file[IC_corona]}"
+        local python_exit_code
+        if __static__Run_Add_Corona_Python_Script; then
+            :
+        else
+            python_exit_code=$?
+            if [[ ${python_exit_code} -eq 3 ]]; then
+                Print_Internal_And_Exit \
+                    'Internal error:' "${target_link_name}" 'found to exist even if it was ensured not to.'
```
RobinSattler: Maybe:

```suggestion
                    'Internal error:' --emph "${target_link_name}" 'found to exist even if it was ensured not to.'
```

RobinSattler (changes requested):
Sorry, for the many comments, but there are several things that should be improved. 
An important thing: the documentation of the new key is missing. On top, I think it would be good to test the add corona behavior in _tests/mocks/smash_afterburner_black-box.py_.

The implementation (I tried the example configs) seems to work fine already! I will check the animation as well, but this will take me probably a couple of days.

I did not check the config settings thoroughly if they are physically meaningful.

configs/predef_configs/config_AuAu_4.3_DynFlu.yaml
RenHirayama: For the test setup it's ok, I think, similar to the other predefined configs.

zpauliny:
> I wonder if we should wait with the merge of this PR until the dynamical fluidization is implemented in vHLLE `main` branch. If I'm not mistaken it's planned to have a config key in vHLLE to choose between hyperbolic and dynamical fluidization runs, right? In this case, we would have to alter the configs you included when the vHLLE work is done if we finish this PR before.

I think the only change relevant for the hybrid is that the *cartesian* keyword is becoming a runtime constant, thus needs to be specified in the config file. This can be easily updated once the new version of vHLLE is tagged.

configs/vhlle_hydro_cartesian
RenHirayama: No because it can be used with the `Constant_Tau` initial conditions, the important thing is that the coordinate system is in cartesian

RenHirayama: i.e. to me, DynFlu is mostly about how one does the core-corona separation

configs/hadron_sampler_cartesian
RenHirayama: The same comment from the vhlle config applies. For the keys, the sampler implementation of $c_s^2$ and $P/\epsilon$ looks wrong to me and I don't know which values they should have, as we change $\epsilon_\mathrm{crit}$.

configs/predef_configs/config_AuAu_4.3_DynFlu.yaml
RenHirayama: Actually it's not clear to me, is the purpose of the predefined configs to offer realistic setups? or just an example that the user should tune?

zpauliny: I think we were shooting for realistic previously.

configs/vhlle_hydro_cartesian
RobinSattler: I see your point in keeping it generic, but is it likely that this will ever be used not as DynFlu but just plain Cartesian?

configs/hadron_sampler_cartesian
RobinSattler: Fine by me. Do users change them a lot or not would be interesting to know. Do you think, we should remove them from the other configs as well then?

configs/predef_configs/config_AuAu_4.3_DynFlu.yaml
RenHirayama: In this case I'll just omit them here as the realistic ones are in the base config (*vhlle_hydro_cartesian*)

configs/hadron_sampler_cartesian
RenHirayama: I think it's unlikely that people change it. @LucasConstantin will at some point check the impact fixed values have compared to getting them from the hydro directly

RenHirayama: that being said, I agree and for now will include them in the config

bash/Afterburner_functionality.bash
```diff
@@ -128,14 +148,30 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Withou
...
+            "${HYBRID_software_input_file[Afterburner]}" \
+            "${HYBRID_software_input_file[IC_corona]}"
+        local python_exit_code
+        if __static__Run_Add_Corona_Python_Script; then
+            :
+        else
+            python_exit_code=$?
+            if [[ ${python_exit_code} -eq 3 ]]; then
+                Print_Internal_And_Exit \
+                    'Internal error:' --emph "${target_link_name}" 'found to exist even if it was ensured not to.'
```
RobinSattler: Just noticed that `'Internal error:'` is unnecessary because `Print_Internal_And_Exit` already prints `INTERNAL:` to the terminal.

```suggestion
                    --emph "${target_link_name}" 'found to exist even if it was ensured not to.'
```

python/add_corona.py
RobinSattler, line 1: Still white space in file. You use VS Code, right? Then you can just use <kbd>Strg</kbd>+<kbd>k</kbd> <kbd>Strg</kbd>+<kbd>x</kbd> after each other and all the white space will be stripped automatically.

python/add_corona.py
```diff
@@ -0,0 +1,179 @@
...
+    # find number of corona events, they should all be the same
+    n_events_corona = find_out_event_number(args.corona_particle_lists[0])
+    corona_particles = gather_corona_particles(args.corona_particle_lists)
+    # find number of sampled events
+    n_events_sampler = find_out_event_number(args.sampled_particle_list)
+
+    read_sampled_and_write_full_particle_list(corona_particles, n_events_corona, 
+                                              n_events_sampler)
+
+    sys.exit(0)
```
RobinSattler: Add empty line at the end to the file.

configs/hadron_sampler_cartesian
```diff
@@ -0,0 +1,9 @@
+surface_file            =DEFAULT=
+output_dir              =DEFAULT=
+number_of_events        1000
+shear                   1 
+bulk                    1
+ecrit                   0.3339
```
RobinSattler: Maybe more of a question for me to understand. Why do you set `ecrit 0.3339`? Is this an interesting "default" value that would be recommendable for users?

configs/hadron_sampler_cartesian
```diff
@@ -0,0 +1,9 @@
...
+hydro_coordinate_system cartesian
```
RobinSattler: Just cosmetics: would be nice to have the key values start at the same "column", so the first seven would need a couple of tabs on top.

```suggestion
surface_file                  =DEFAULT=
output_dir                    =DEFAULT=
number_of_events              1000
shear                         1 
bulk                          1
ecrit                         0.3339
hydro_coordinate_system       cartesian
```

docs/user/CHANGELOG/index.md
```diff
@@ -48,11 +48,12 @@ Given a version number `X.Y.Z`,
 
     **Changes:**
 
+    :new: &nbsp; Added `Add_corona_from_IC_and_Hydro` configuration key in the `Afterburner` module, which merges the files containing corona particles from :file_folder: ***IC*** and :file_folder: ***Hydro*** into the particle lists sampled in :file_folder: ***Sampler***.
```
RobinSattler, line 51: Minor suggestion:

```suggestion
    :new: &nbsp; Added `Add_corona_from_IC_and_Hydro` configuration key in the `Afterburner` module, which merges the files containing corona particles from :file_folder: ***IC*** and :file_folder: ***Hydro*** into the particle lists sampled in the :file_folder: ***Sampler*** stage.
```

python/add_corona.py
```diff
@@ -0,0 +1,179 @@
...
+#===================================================
+
+import argparse
+import numpy as np
+import sys
+import os
+import subprocess
+
+'''
+    This script adds the corona particles from the initial SMASH run
```
RobinSattler: Naively, this sounds to me like there's only one list from the IC stage. Maybe add that this includes a list of particles during the Hydro stage that didn't undergo fluidization due to the live time of the fluid (if I understood this right). But this might be already to verbose, so just a suggestion that you can use or disregard at will.

python/add_corona.py
```diff
@@ -0,0 +1,179 @@
...
+    '''
+        Look into the last line of the file and find the number of events
+    '''
+    if not os.path.isfile(filepath):
+        print(filepath+" does not exist!")
+        sys.exit(3)
+    tail = subprocess.run(['tail', '-n', '1', filepath],
+                          stdout=subprocess.PIPE,stderr=subprocess.PIPE)
+    last_line = tail.stdout.decode().strip()
+    if "end" in last_line:
```
RobinSattler, line 35: Since this was exactly a bug in SMASH recently, I'd be in favor of
```suggestion
    if " end " in last_line:
```
which makes it slightly safer.

configs/predef_configs/config_AuAu_4.3_DynFlu.yaml
```diff
@@ -0,0 +1,38 @@
...
+                    Particles: {2212: 79, 2112: 118} #Gold197
+                Sqrtsnn: 4.3
+        Impact:
+            Range: [0.0,3.3] # 0-5%
+        Initial_Conditions:
+            Type: "Dynamic"
+            Energy_Density_Threshold: 0.5
+            Formation_Time_Fraction: 0.2
+            Fluidizable_Processes: ["All"]
+            Delay_Initial_Elastic: true
```
RobinSattler: `Impact` and `Initial_Conditions` are part of the `Collider` section, so there's indentation missing.

RobinSattler (commented):
I started reviewing and then I noticed that you probably are not done yet &mdash; since the commit message hints at it and some changes lead to a crash (see my comment in the handler config file). Since I already found some things, I leave them here as general feedback and will review further when you give me green light for it.

configs/hadron_sampler_cartesian
RenHirayama: this is the value found by the bayesian analysis

docs/user/CHANGELOG/index.md
RenHirayama: I disagree because the statement is about the folder, not the stage necessarily

python/add_corona.py
RenHirayama: i put generically "and possibly others" as there is no specific mention of the other corona coming from hydro (nor it needs to be enforced)

python/add_corona.py
RenHirayama: actually I changed the line to be `split()`, i.e. its an array that needs to contain the element `"end"`, instead of the full string

RobinSattler: I like it! :+1:

bash/utility_functions.bash
```diff
@@ -669,7 +669,8 @@ function Make_Functions_Defined_In_This_File_Readonly()
     #
     # NOTE: The file from which this function is called is ${BASH_SOURCE[1]}
     local declared_functions
-    declared_functions=( # Here word splitting can split names, no space allowed in function name!
+    # Here word splitting can split names, no space allowed in function name!
```
RenHirayama, line 672: This kept failing with my local formatter for some reason and anyhow its better to put the comment outside

configs/predef_configs/config_AuAu_4.3_DynFlu.yaml
```diff
@@ -0,0 +1,38 @@
...
+                    Particles: {2212: 79, 2112: 118} #Gold197
+                Sqrtsnn: 4.3
+            Impact:
+                Range: [0.0,3.3] # 0-5%
+            Initial_Conditions:
+                Type: "Dynamic"
+                Energy_Density_Threshold: 0.5
+                Formation_Time_Fraction: 0.2
+                Fluidizable_Processes: ["All"]
+                Delay_Initial_Elastic: true
```
RobinSattler: Test runs crash since indentation wrong.

```suggestion
                Impact:
                    Range: [0.0,3.3] # 0-5%
                Initial_Conditions:
                    Type: "Dynamic"
                    Energy_Density_Threshold: 0.5 
                    Formation_Time_Fraction: 0.2 
                    Fluidizable_Processes: ["All"]
                    Delay_Initial_Elastic: true
```

docs/user/configuration_file.md
```diff
@@ -229,6 +229,10 @@ Sampler:
     However, using this key any file path can be specified.
     This key is ignored if `Add_spectators_from_IC` is set to `false`.
 
+???+ config-key "`Add_corona_from_IC_and_Hydro`"
+
+    Append the corona particles from the `IC` and `Hydro` stages to the particles sampled in the `Sampler` in an event using dynamic fluidization. If `true`, a :material-file: *particle_lists.oscar* file is expected to exist in the :file_folder: ***IC*** output sub-folder with the same `Run_ID`. The :file_folder: ***Hydro*** folder may contain it or not.
```
RobinSattler: I think a note about the mutually exclusiveness with the `Add_spectators_from_IC` key is important and this should be included in the sanity checks, i.e. the handler should throw when both keys are set to `true`. You should also state the default value.

```suggestion
    Append the corona particles from the `IC` and `Hydro` stages to the particles sampled in the `Sampler` in an event using dynamic fluidization. If `true`, a :material-file: *particle_lists.oscar* file is expected to exist in the :file_folder: ***IC*** output sub-folder with the same `Run_ID`. The :file_folder: ***Hydro*** folder may contain such a file as well. The default value is `false`.

    ⚠️ This key is mutually exclusive with the `Add_spectators_from_IC` key. If both are set to `true`, it will result in an error.
```

bash/Afterburner_functionality.bash
```diff
@@ -128,14 +148,38 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Withou
...
             fi
         fi
-    else
-        if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
-            ln -s -f "${HYBRID_software_input_file[Afterburner]}" "${target_link_name}"
-        elif [[ ! "${target_link_name}" -ef "${HYBRID_software_input_file[Afterburner]}" ]]; then
-            exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
-                'File ' --emph "${target_link_name}" ' exists but it is not the Afterburner input file ' \
-                --emph "${HYBRID_software_input_file[Afterburner]}" ' to be used.'
+    elif [[ "${HYBRID_optional_feature[Add_corona_from_IC_and_Hydro]}" = 'TRUE' ]]; then
```
RobinSattler: If `Add_spectators_from_IC` is not explicitly set to `false` this will never be entered since the default is `true`. I think this is suboptimal from the user perspective and both keys being `true` should result in an error in sanity checks. The minimum is to stress it in the documentation.

python/add_corona.py
RobinSattler: Just a suggestion, but why not more verbose and precise like:


```suggestion
    This script adds the corona particles from the IC and possibly
    the Hydro stage to the sampled particles list.
```

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+    particles = np.empty(shape=[0, 13])
+    n_event = 0
+
+    for line in open(filename, "r"):
+        line = line.split()
+        # event end line
+        if "end" in line:
+            n_event += 1
+        # ignore header
+        elif "#" in line or "#!OSCAR2013" in line or "#!OSCAR2013Extended" in line:
```
RobinSattler: This is just to skip the file and the event headers, right? What about using the following instead? This will be more robust to changes to the `particle_lists.oscar` output files.

```suggestion
        elif len(line) == 0 or line[0].startswith("#"):
```

This check also skips empty lines as a safety feature. Otherwise, the handler will crash with the not so verbose `FATAL: Adding corona from IC and Hydro to the sampled particles file failed.` if the `particle_lists.oscar` file contains any empty lines (although unlikely).

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+            continue
+        # particle line
+        else:
+            particle = np.array(line[:12])
+            particle = np.append(int(n_event), particle)
```
RobinSattler: I'm not an expert on this and you can go with what you like better, but wouldn't unpacking the list be an easier way of dealing with it?

```suggestion
            particle = np.array([int(n_event), *line[:12]])
```

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+    return n_events+1
+
+def extract_particles(filename):
+    '''
+        Extract particles from a file to a numpy array
+    '''
+    particles = np.empty(shape=[0, 13])
+    n_event = 0
+
+    for line in open(filename, "r"):
```
RobinSattler: If this opening method is chosen, the file should be closed manually if I'm not mistaken. You could use
```suggestion
    with open(filename, "r") as f:
        for line in f:
```
instead, which closes the file automatically.

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+        Corona particles are distributed among sampled events
+    '''
+    output_file = args.output_file
+    copy_header(args.sampled_particle_list,output_file)
+
+    sampled_particles = np.empty(shape=[0, 12])
+    event_s = 0
+    event_c = 0
+
+    for line in open(args.sampled_particle_list, "r"):
```
RobinSattler: Again, this should be closed manually or you use the `with open(filename, "r") as f` method.

And it would be good if you check beforehand if the `sampled_particles.oscar` file exists and if not exit with `sys.exit(2)`. Otherwise, lines 162-164 in _bash/Afterburner_functionality.bash_ are not accurate because you never checked if the Afterburner input file exists or not.

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+                print("read and write event "+str(event_s))
+            # reset relevant variables
+            if event_c < (n_events_ic-1):
+                event_c += 1
+            else:
+                event_c = 0
+            event_s += 1
+            sampled_particles = np.empty(shape=[0, 12])
+        # ignore event headers
+        elif "#" in line[0]:
```
RobinSattler: I'd use here the same safety check to skip empty lines, to avoid an `IndexError`.
```suggestion
        elif len(line) == 0 or "#" in line[0]:
```

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-s","--sampled_particle_list", required = True,
+                        help="File containing the sampled particle lists.")
+    parser.add_argument("-c","--corona_particle_lists", nargs='+', required = True,
+                        help="Particle list from the initial conditions SMASH run.")
+    parser.add_argument("-o","--output_file", required = True,
+                        help="Resulting particle list containing " \
+                        "sampled and spectator particles.")
+    parser.add_argument("-f", "--force", action='store_true',
+                        help = "Ignore pre-existing output file")
```
RobinSattler:
```suggestion
                        help = "Ignore pre-existing output file.")
```

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+        print("No corona particles!", file=sys.stderr)
+        sys.exit(4)
+    if not is_any_corona_file_present:
+        print("Corona files not found!", file=sys.stderr)
+        sys.exit(2)
+    return corona_particles
+
+def copy_header(input_file, output_file):
+    header = ""
+    for line in open(input_file, "r"):
```
RobinSattler: Again, this should be closed manually or you use the `with open(filename, "r") as f` method.

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+        if line.startswith('# event'):
```
RobinSattler: I think this adds additional safety if the user toys around with `particle_lists.oscar` files in between running stages.
```suggestion
        if line.startswith('# event') or not line.startswith('#'):
```

bash/Afterburner_functionality.bash
```diff
@@ -128,14 +148,38 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Withou
...
+        else
+            python_exit_code=$?
+            if [[ ${python_exit_code} -eq 2 ]]; then
+                Print_Internal_And_Exit \
+                    --emph "${HYBRID_software_input_file[IC_corona]}" \
+                    'or' --emph "${HYBRID_software_input_file[Afterburner]}" \
+                    'not found even if it was ensured to exist.'
+            elif [[ ${python_exit_code} -eq 3 ]]; then
+                Print_Fatal_And_Exit \
+                    'Could not determine number of events from last line of some input file.'
```
RobinSattler: Only a suggestion, choose at your will.
```suggestion
                    'Could not determine number of events from last line of at least one input file.'
```

bash/Afterburner_functionality.bash
```diff
@@ -128,14 +148,38 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link_With_Or_Withou
...
+            elif [[ ${python_exit_code} -eq 4 ]]; then
+                Print_Warning \
+                    'No corona particles were found, making a symbolic link for the sampled particles.'
```
RobinSattler:
```suggestion
                    'No corona particles were found, creating a symbolic link for the sampled particles.'
```

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+        Extract particle lines from the initial conditions oscar output of SMASH
+        and any extra file (e.g. from vHLLE oscar output) and join particles
+        into one array.
+    '''
+    is_any_corona_file_present = False
+    corona_particles = np.empty(shape=[0, 13])
+    for file in corona_lists:
+        if os.path.isfile(file):
+            is_any_corona_file_present = True
+            more_corona_particles = extract_particles(file)
```
RobinSattler: Minor suggestion:
```suggestion
            corona_particles_in_file = extract_particles(file)
```

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+            corona_particles = np.append(corona_particles, more_corona_particles, axis=0)
+    if len(corona_particles) == 0:
+        print("No corona particles!", file=sys.stderr)
```
RobinSattler:
```suggestion
        print("No corona particles found!", file=sys.stderr)
```

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+        sys.exit(4)
+    if not is_any_corona_file_present:
```
RobinSattler, line 82: This `if not is_any_corona_file_present:` statement has to be before the check `if len(corona_particles) == 0:`, otherwise it will never be entered. You create an empty array in line 77 and if no corona file is found the array is still empty and the `if len(corona_particles) == 0:` will always be true.

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+            with open(output_file, 'a') as f:
+                np.savetxt(f, sampled_particles, delimiter=' ', fmt='%s')
+            # write corona
+            with open(output_file, 'a') as f:
+                np.savetxt(f, corona_filter[:,1:], delimiter=' ', fmt='%s')
+            with open(output_file, 'a') as f:
+                f.write("# event {} end \n".format(event_s))
+
+            if event_s % 100 == 0:
+                print("read and write event "+str(event_s))
```
RobinSattler: This is printed to standard output during the handler runs and I don't understand yet why.

python/add_corona.py
```diff
@@ -0,0 +1,173 @@
...
+            with open(output_file, 'a') as f:
+                f.write("# event {} out {}\n".format(event_s,particle_number))
+            # write sampled
+            with open(output_file, 'a') as f:
+                np.savetxt(f, sampled_particles, delimiter=' ', fmt='%s')
+            # write corona
+            with open(output_file, 'a') as f:
+                np.savetxt(f, corona_filter[:,1:], delimiter=' ', fmt='%s')
+            with open(output_file, 'a') as f:
+                f.write("# event {} end \n".format(event_s))
```
RobinSattler:
```suggestion
                f.write("# event {} end\n".format(event_s))
```

tests/functional_tests_Afterburner_only.bash
```diff
@@ -192,6 +200,7 @@ function Functional_Test__do-Afterburner-only()
             File_Directory: "."
     ' "${run_id}" "${mocks_folder}" > "${config_filename}"
     Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
+    __static__Check_Existence_Sampled_Particles "Afterburner/${run_id}"
     __static__Check_Successful_Handler_Run $?
```
RobinSattler: Your new function should come after the function `__static__Check_Successful_Handler_Run $?`. Otherwise, you're killing the purpose of this function.

```suggestion
    __static__Check_Successful_Handler_Run $?
    __static__Check_Existence_Sampled_Particles "Afterburner/${run_id}"

```

tests/functional_tests_Afterburner_only.bash
```diff
@@ -246,6 +255,7 @@ function Functional_Test__do-Afterburner-only()
             File_Directory: "."
     ' "${run_id}" "${mocks_folder}" "$(pwd)" > "${config_filename}"
     Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
+    __static__Check_Existence_Sampled_Particles "Afterburner/${run_id}"
     __static__Check_Successful_Handler_Run $? || return 1
```
RobinSattler: Same here: your new function should come after the function `__static__Check_Successful_Handler_Run $?`. Otherwise, you're killing the purpose of this function.

tests/functional_tests_Afterburner_only.bash
```diff
@@ -16,6 +16,14 @@ function __static__Check_Successful_Handler_Run()
     Check_If_Software_Produced_Expected_Output 'Afterburner' "$(pwd)/Afterburner"
 }
 
+function __static__Check_Existence_Sampled_Particles()
+{
+    if [ ! -f "$1/sampled_particles.oscar" ]; then
+        exit_code=1 Print_Fatal_And_Exit 'Hybrid-handler unexpectedly failed.'
+        return 1
+    fi
+}
```
RobinSattler, line 26: Why is there a need for this function? The function above `__static__Check_Successful_Handler_Run` already checks with `Check_If_Software_Produced_Expected_Output 'Afterburner' "$(pwd)/Afterburner"` if everything is in place (at least number of files wise).

If we want to keep this, the error message should be more verbose, e.g.,
```bash
        exit_code=1 Print_Fatal_And_Exit 'The ' --emph "sampled_particles.oscar" ' file is expected to exist, but was not found.'
```

tests/functional_tests_Afterburner_only.bash
```diff
@@ -288,4 +298,56 @@ function Functional_Test__do-Afterburner-only()
...
+    Afterburner:
+      Executable: %s/smash_afterburner_black-box.py
+      Add_spectators_from_IC: false
+      Add_corona_from_IC_and_Hydro: true
+      Software_keys:
+        Modi:
+          List:
+            File_Directory: "."
+    ' "${run_id}" "${mocks_folder}" > "${config_filename}"
+    Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
```
RobinSattler, line 342: Add a check if the run was successful.
```suggestion
    Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
    __static__Check_Successful_Handler_Run $?
```

RobinSattler (changes requested):
I think the most important part to add is a (sanity) check for the exclusiveness of the `Add_spectators_from_IC` and the `Add_corona_from_IC_and_Hydro` key. At the moment, there is no safety feature in place and this will only lead to the running of the `add_spectators.py` script if both keys are set to `true` &mdash; since the default for `Add_spectators_from_IC` is `true` this is not great. It might be also a good option to enforce `Add_spectators_from_IC: false` if `Add_corona_from_IC_and_Hydro: true`.

Additionally, I'm not quite sure if a unit test for the add corona behavior in _tests/unit_tests_Afterburner_functionality.bash_ would be good to have as well.

Since the code review took already a lot of time &mdash; and will take up some more with the next iteration &mdash; and I'm not really familiar with the proper dynamical fluidization settings, it would be great if someone else could check the added configs regarding physics correctness. @zpauliny, do you have capacity to check them?

tests/functional_tests_Afterburner_only.bash
RenHirayama: Because checking for the specific input file (whose name is assumed in the codebase) is better than just the number of files produced.

tests/functional_tests_Afterburner_only.bash
```diff
@@ -16,6 +16,15 @@ function __static__Check_Successful_Handler_Run()
     Check_If_Software_Produced_Expected_Output 'Afterburner' "$(pwd)/Afterburner"
 }
 
+function __static__Check_Existence_Sampled_Particles()
+{
+    if [ ! -f "$1/sampled_particles.oscar" ]; then
```
RobinSattler:
```suggestion
    if [[ ! -f "$1/sampled_particles.oscar" ]]; then
```

bash/Afterburner_functionality.bash
```diff
@@ -90,10 +90,62 @@ function __static__Run_Add_Spectators_Python_Script()
...
     local -r target_link_name="${HYBRID_software_output_directory[Afterburner]}/${HYBRID_afterburner_list_filename}"
-    if [[ "${HYBRID_optional_feature[Add_spectators_from_IC]}" = 'TRUE' ]]; then
+    if [[ "${HYBRID_optional_feature[Add_corona_from_IC_and_Hydro]}" = 'TRUE' ]]; then
+        Ensure_Given_Files_Do_Not_Exist "${target_link_name}"
+        Ensure_Given_Files_Exist \
+            "${HYBRID_software_input_file[Afterburner]}" \
+            "${HYBRID_software_input_file[IC_corona]}"
+        local python_exit_code
+        if __static__Run_Add_Corona_Python_Script; then
+            :
```
RobinSattler, line 126: I think the comment we have for the same line for the spectators script makes sense here.
```suggestion
        if __static__Run_Add_Corona_Python_Script; then
            # BE AWARE: The negation of the if-clause does not work because then the exit code cannot
            #           be extracted properly (it will always be 0 because of the true if-statement)
            :
```

bash/Afterburner_functionality.bash
```diff
@@ -90,10 +90,62 @@ function __static__Run_Add_Spectators_Python_Script()
...
+            elif [[ ${python_exit_code} -eq 5 ]]; then
+                Print_Internal_And_Exit \
+                    --emph "${target_link_name}" 'already exists even if it was ensured not to.'
+            else
+                Print_Fatal_And_Exit \
+                    'Adding corona from IC and Hydro to the sampled particles file failed.'
+            fi
+        fi
+    # Right now, the order of checking for corona and then spectators allows for the user to only set the former
+    # to `true`, without having to set the latter to `false` (the default behavior is to include spectators)
```
RobinSattler: This comment is not true at the moment since the exclusiveness check in _bash/sanity_checks.bash_ leads to a crash when both are set to `true`.

bash/sanity_checks.bash
```diff
@@ -209,6 +211,16 @@ function __static__Perform_Logic_Checks_Depending_On_Execution_Mode()
     esac
 }
 
+function __static__Perform_Logic_Checks_For_Exclusive_Options()
+{
+    if [[ "${HYBRID_optional_feature[Add_corona_from_IC_and_Hydro]}" = 'TRUE' ]] \
+        && [[ "${HYBRID_optional_feature[Add_spectators_from_IC]}" = 'TRUE' ]]; then
+        exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
+            'The Afterburner keys ' --emph 'Add_spectators_from_IC' ' and ' --emph 'Add_corona_from_IC_and_Hydro' \
+            ' cannot both be true simultaneously.'
```
RobinSattler:
```suggestion
            ' cannot both be set to true simultaneously.'
```

bash/sanity_checks.bash
```diff
@@ -330,6 +342,33 @@ function __static__Set_Software_Input_Data_File()
...
+                    exit_code=${HYBRID_fatal_logic_error} Print_Fatal_And_Exit \
+                        'Requesting custom ' --emph 'Hydro_corona' ' input file although executing ' \
+                        --emph 'Hydro' ' with default output name.'
+                fi
+            else
+                printf -v HYBRID_software_input_file[Hydro_corona] '%s/%s' \
+                    "${HYBRID_software_output_directory[Hydro]}" \
+                    "${HYBRID_software_default_input_filename[Hydro_corona]}"
+            fi
+        fi
```
RobinSattler, line 371: In principle, since these two checks for IC and Hydro are very similar, the code could be reduced by a function. But I think it's not strictly necessary here.

tests/functional_tests_Afterburner_only.bash
```diff
@@ -175,6 +184,31 @@ function Functional_Test__do-Afterburner-only()
             --emph "${HYBRID_fatal_logic_error}" '.'
         return 1
     fi
+    # Expect failure if both corona and spectators are epected
+    Print_Info 'Running Hybrid-handler expecting failure from requesting' \
+        'both corona and spectators simultaneously.'
```
RobinSattler:
```suggestion
        'to add both corona and spectators simultaneously.'
```

tests/functional_tests_Afterburner_only.bash
```diff
@@ -175,6 +184,31 @@ function Functional_Test__do-Afterburner-only()
...
+      Add_spectators_from_IC: true
+      Add_corona_from_IC_and_Hydro: true
+      Software_keys:
+        Modi:
+          List:
+            File_Directory: "."
+    ' "${run_id}" "${mocks_folder}" > "${config_filename}"
+    Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
+    local res=$?
+    echo "res = ${res}"
```
RobinSattler: What is this line for? :thinking:

tests/functional_tests_Afterburner_only.bash
```diff
@@ -175,6 +184,31 @@ function Functional_Test__do-Afterburner-only()
...
+    if [[ ${res} -ne ${HYBRID_fatal_logic_error} ]]; then
```
RobinSattler: I'd prefer:

```suggestion
    if [[ $? -ne ${HYBRID_fatal_logic_error} ]]; then
```
instead of storing it in a variable first.

RobinSattler (commented):
All in all it looks quite good now and I like most of the implemented changes! 👍 Nevertheless, on top of a few minor comments, there are still a couple things that should be taken care of:
- One thing that still bugs me is that if I only set `Add_corona_from_IC_and_Hydro: true` in the hybrid config file, the run will crash because of the exclusiveness check (which makes total sense). I think if users set this key deliberately to `true`, they should not be forced to put `Add_spectators_from_IC: false`. Either we should make `Add_spectators_from_IC: false` the default behavior or setting `Add_corona_from_IC_and_Hydro: true` should automatically set `Add_spectators_from_IC: false`. I like the change of the `Add_spectators_from_IC` default behavior better. I think this decision should be taken before merging this PR.
- Somehow there is a commit duplicate in your branch history now which should be removed. Commit [4676085](https://github.com/smash-transport/smash-vhlle-hybrid/pull/86/commits/4676085d42d275e491f7970a02061544ad472f43) has to be dropped.
- An issue should be opened about the unit tests (for python and _tests/unit_tests_Afterburner_functionality.bash_ should be checked if a unit test makes sense in there too).

tests/functional_tests_Afterburner_only.bash
```diff
@@ -193,6 +227,7 @@ function Functional_Test__do-Afterburner-only()
     ' "${run_id}" "${mocks_folder}" > "${config_filename}"
     Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
     __static__Check_Successful_Handler_Run $?
+    __static__Check_Existence_Sampled_Particles "Afterburner/${run_id}"
```
RobinSattler, line 227: In my opinion, these two functions should always have a trailing `|| return 1` (in the entire file) as a safety feature for the function call. Since I'm not entirely sure about this, I asked Alessandro and am waiting for his reply.

```suggestion
    __static__Check_Successful_Handler_Run $? || return 1
    __static__Check_Existence_Sampled_Particles "Afterburner/${run_id}" || return 1
```

tests/functional_tests_Afterburner_only.bash
RenHirayama: why?

bash/sanity_checks.bash
RenHirayama: then I won't do it

bash/Afterburner_functionality.bash
RenHirayama: this comment tells me nothing tbh

tests/functional_tests_Afterburner_only.bash
RenHirayama: it works without it, so i don't understand what the point of adding it

RenHirayama:
> I like the change of the Add_spectators_from_IC default behavior better. I think this decision should be taken before merging this PR

As per the mattermost discussion, this should stay `true` by default.

tests/functional_tests_Afterburner_only.bash
RobinSattler: Codebase convention since `[[ ]]` has some extra features, more info [here](https://unix.stackexchange.com/questions/49007/when-should-i-use-vs-in-bash-single-vs-double-brackets).

bash/Afterburner_functionality.bash
RobinSattler: One might think that negating the `if` statement will shorten code since you don't need the `else` anymore. This does not work since then the exit code cannot be extracted, which is what the comment states.

tests/functional_tests_Afterburner_only.bash
RobinSattler: Alessandro is in favor of leaving it out. So it should be removed everywhere.

tests/functional_tests_Afterburner_only.bash
```diff
@@ -273,6 +307,7 @@ function Functional_Test__do-Afterburner-only()
             --emph "${HYBRID_fatal_logic_error}" '.'
         return 1
     fi
+    mv 'Afterburner' 'Afterburner-failure-combine-IC-and-custom-spectators'
```
RobinSattler: As discussed in person, this has to be removed since the folder is never created due to the early failure.

tests/functional_tests_Afterburner_only.bash
```diff
@@ -175,6 +184,29 @@ function Functional_Test__do-Afterburner-only()
...
+            File_Directory: "."
+    ' "${run_id}" "${mocks_folder}" > "${config_filename}"
+    Run_Hybrid_Handler_With_Given_Options_In_Subshell 'do' '-c' "${config_filename}" '-o' '.'
+    if [[ $? -ne ${HYBRID_fatal_logic_error} ]]; then
+        Print_Error \
+            'Hybrid-handler did not fail as expected with exit code ' \
+            --emph "${HYBRID_fatal_logic_error}" '.'
+        return 1
+    fi
+    mv 'Afterburner' 'Afterburner-failure-both-corona-and-spectator'
```
RobinSattler: As discussed in person, this has to be removed since the folder is never created due to the early failure.

bash/Afterburner_functionality.bash
```diff
@@ -90,10 +90,60 @@ function __static__Run_Add_Spectators_Python_Script()
...
     local -r target_link_name="${HYBRID_software_output_directory[Afterburner]}/${HYBRID_afterburner_list_filename}"
-    if [[ "${HYBRID_optional_feature[Add_spectators_from_IC]}" = 'TRUE' ]]; then
+    if [[ "${HYBRID_optional_feature[Add_corona_from_IC_and_Hydro]}" = 'TRUE' ]]; then
+        Ensure_Given_Files_Do_Not_Exist "${target_link_name}"
+        Ensure_Given_Files_Exist \
+            "${HYBRID_software_input_file[Afterburner]}" \
+            "${HYBRID_software_input_file[IC_corona]}"
+        local python_exit_code
+        if __static__Run_Add_Corona_Python_Script; then
+            :
```
RobinSattler, line 126: I'd still be in favor of the comment here, but leave it up to you.

bash/sanity_checks.bash
RobinSattler: This got somehow lost in your newest push. :thinking:

RobinSattler (approved):
Thanks for all the changes! This looks sound now! 🚀 I left minor comments, but mostly only suggestions you can choose or leave out.

PR #90, opened 2025-11-24
Add missing documentation for integration tests and update authors

RobinSattler:
I noticed that we forgot to add documentation about the integration tests. On top, I removed Niklas' affiliation and ORCID since he's not an active developer anymore.


docs/developer/testing_framework.md
```diff
@@ -119,6 +121,14 @@ If such a function exists, the runner will check for existence of the correspond
     }
     ```
 
+=== "Integration test function"
```
AxelKrypton: In the documentation, this box is having now many windows with long titles and they do not fit anymore in the available horizontal space leading to a horizontal scrolling. I suggest to remove the word `function` from all of them, changing `Setup|Teardown function` to `Test setup|teardown` or simply `Setup` and `Teardown` as you prefer.

docs/developer/testing_framework.md
```diff
@@ -137,7 +147,9 @@ If such a function exists, the runner will check for existence of the correspond
 
 !!! tip "Use the framework functionality to invoke code to test"
     * Codebase functions to be invoked in unit tests should be called through the `Call_Codebase_Function` and `Call_Codebase_Function_In_Subshell` interface functions (passing the name of the function to be invoked as first argument and the arguments to be forward afterwards).
-    *  In functional tests you'll probably want to run the hybrid handler with some options and this can be easily achieved by using the `Run_Hybrid_Handler_With_Given_Options_In_Subshell` function.
+    * Functions in integration tests should be called with the `Call_Codebase_Function` interface functions.
```
AxelKrypton: A possible more complete alternative (adjust English as you wish):

```suggestion
    * Functions in integration tests should be called with the `Call_Codebase_Function` interface functions.
      An integration test is about testing different functionality together and, in this sense, side effects of a function call like e.g. variables change should persist beyond the call (this would not be the case if the function is run in a sub-shell), because such tests usually are a chain of function calls.
      Therefore, it is often the developer responsibility to manually introduce a sub-shell `(...)` around the integration, if this is desired.
```

AxelKrypton (approved):
I have two minor suggestions, adapt according to your taste before merging. 🟢

Issue #87, opened 2025-11-17
Update SMASH output format from `ASCII` to `For_vHLLE` + adjust `global_variables.bash`


LucasConstantin:
**Description:**
Since the `ASCII` output format has been replaced by the new `For_vHLLE` format in recent SMASH versions, several configuration files in this repository no longer work as expected.

### Affected configuration

For example, in
`configs/smash_initial_conditions__ge_v3.2.yaml`
the output format is still set to `ASCII`:
[https://github.com/smash-transport/smash-vhlle-hybrid/blob/18ca3a95eec3e69c1a0346e5cbac1074e4a944f3/configs/smash_initial_conditions__ge_v3.2.yaml#L14](https://github.com/smash-transport/smash-vhlle-hybrid/blob/18ca3a95eec3e69c1a0346e5cbac1074e4a944f3/configs/smash_initial_conditions__ge_v3.2.yaml#L14)

With SMASH ≥ 3.3, the output file is now named:

```
SMASH_IC_For_vHLLE.dat
```

### Required script update

Because of the new file name, the path defined in
`bash/global_variables.bash`
also needs to be updated:
[https://github.com/smash-transport/smash-vhlle-hybrid/blob/18ca3a95eec3e69c1a0346e5cbac1074e4a944f3/bash/global_variables.bash#L40](https://github.com/smash-transport/smash-vhlle-hybrid/blob/18ca3a95eec3e69c1a0346e5cbac1074e4a944f3/bash/global_variables.bash#L40)

### Request for review

I am not fully certain whether additional changes are required elsewhere in the workflow. It would be great if others familiar with the hybrid setup could check for any other adjustments needed due to the format change.

Thanks!

---

RobinSattler:
This looks like we need a new base config for the IC of `SMASH-3.3` and the proper implementation of all the changes will require the use of the "version logic" that was implemented in #61 and #69 to stay backward compatible. @LucasConstantin, I can assist by pointing you in the right direction, but I do not have enough capacity to take care of this at the moment. A good starting point would be to `grep` for all `SMASH_IC.dat` and `ASCII` occurrences in the codebase, and to check how we distinguish between the IC configs of `SMASH-3.2` and all the other public versions, i.e. how the "version logic" works.

@RenHirayama, did anything else change apart from the IC input name? The `SMASH_IC.oscar` stayed the same?

AxelKrypton:
[...]

PR #92, opened 2025-11-28
Implement smarter software versions handling

AxelKrypton:
This PR is meant to help @LucasConstantin in dealing with (part of) #87.

In general, it should be now trivial to add new base configuration files for new software versions that introduce some breaking change. For the `Hydro` and the `Afterburner` stages there was not yet a need to select the base configuration depending on the version and I did not introduce it. For the `Afterburner` stage it would be trivial to do (it would be 1:1 to the IC), but for hydro it would be difficult as there is no way to ask `vHLLE` for its version.

I took this opportunity to tidy up the ***configs*** folder, as @RenHirayama introduced there configuration files which are not used as base by the handler. 😅 Predefined configs are in the dedicated subfolder. I added now a README there.

@RobinSattler I assigned you here as you are familiar with the selecting mechanism we implemented together back in time. Feel free to reassign if this does not fit your schedule. @LucasConstantin You could also try to review these changes to get more familiar with the codebase. If some Bash sounds strange, feel free to ask.


RobinSattler:
@AxelKrypton, as discussed on Mattermost, I will not find time right now to do the review. Because of his upcoming work, I think it makes sense to re-assign to @LucasConstantin.

tests/utility_functions.bash
LucasConstantin, line 1: What is the point of this?

bash/sanity_checks.bash
```diff
@@ -134,36 +134,34 @@ function __static__Set_Base_Configuration_File_If_Unset()
...
             Print_Internal_And_Exit 'Base configuration file unset for ' --emph "${key}" \
                 '\nstage, although this should not be the case!'
             ;;
     esac
+    readonly prefix extension
+    # If path is still unset, it means the decision is to be taken based on software version
+    if [[ ${path} = '' ]]; then
+        Ensure_That_Given_Variables_Are_Set_And_Not_Empty "HYBRID_software_version[${key}]"
+        path="$(__static__Pick_Base_Config_To_Be_Used \
+            "${key}" "${prefix}" "${extension}" "${HYBRID_software_version[${key}]}")"
```
LucasConstantin, line 162: Why is this the correct order for prefix, extension and version number? Shouldnt the extension be last?

LucasConstantin (approved):
These changes seem reasonable to me, I just left some remaining questions as comments.

bash/sanity_checks.bash
AxelKrypton: There is no _correct_ order, in the sense that the four arguments passed to the functions just need to respect the "function interface" i.e. how the functions interprets them. This is found in the function implementation, that is in this same file on lines 441 on. In particular, there you find
```bash
    local -r \
        key=$1 \
        prefix=$2 \
        extension=$3 \
        software_version=$4
```
and this is how the function expects them. In Bash it is up to the developer to choose and ensure an interface. One could argue that a check on the number of arguments `$#` passed to the function might be added, true, but this is a utility function and we can also argue that we are in a place where the developer knows what should be done. And for consistency this is an assumption we made for basically all utility `__static__*` functions.

tests/utility_functions.bash
AxelKrypton: I guess you mean the point of this change, not of the file (although you left a comment on the file). I moved that pretty frequent printing on the `TRACE` level of the logger because I needed to run the test with debug output and it was too much polluted by this information, which is very low level. Just a change to improve usefulness of debug output keeping it a bit cleaner than it was.

PR #93, opened 2025-12-09
Sass/add spin keys

nilssass:
This PR prepares the handler for the spin setup in SMASH 3.3 and the next release of the sampler. I will open the PR already such that it can be merged after the next SMASH release.

The new keys do not break any code even before the corresponding merges of the spin parts as the sampler and vHLLE simply ignore unknown keys with a warning message.


bash/Sampler_functionality_SMASH.bash
```diff
@@ -45,6 +45,8 @@ function Validate_Configuration_File_Of_SMASH()
         'ecrit'
         'cs2'
         'ratio_pressure_energydensity'
+        'sample_spin'
+        'vorticity_vector'
```
RobinSattler, line 49: These should be included in the part where we use version logic since the new keys are only available from Sampler version `3.3` on.

configs/hadron_sampler__ge_v3.2
```diff
@@ -6,5 +6,7 @@ bulk                          0
 shear                         1
 cs2                           0.15
 ratio_pressure_energydensity  0.15
+sample_spin                   0
+vorticity_vector              0
 create_root_output            0
 hydro_coordinate_system       tau-eta
```
RobinSattler, line 12: Not the way to go since this config can be used for version `3.2` which does not incorporate the keys. The new keys (including `hydro_coordinate_system`) should only be available from Sampler version `3.3` on.

This will even become crucial for Sampler versions from `3.3` on since Ren implemented that the Sampler will throw if an unknown key is present.

bash/Sampler_functionality_SMASH.bash
```diff
@@ -45,6 +45,8 @@ function Validate_Configuration_File_Of_SMASH()
...
     )
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
         allowed_keys+=('create_root_output' 'hydro_coordinate_system')
```
RobinSattler, line 52: I just realized that I made a mistake here and the key `hydro_coordinate_system` should only be available from Sampler version `3.3` on. Would be good if you could fix this in the course of this PR. I'm not entirely sure this is the way to go, but one could think about the following:
```suggestion
    if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
        allowed_keys+=('create_root_output')
    fi
    if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.3'
        allowed_keys+=('hydro_coordinate_system' 'sample_spin' 'vorticity_vector')
```

RobinSattler (changes requested):
You need to use the version logic for your new keys. I left a suggestion, but you need to check if this is the best way to go.

nilssass:
@AxelKrypton In which order should we merge the PRs? I can also include the suggestions by @RobinSattler and merge this PR first.

AxelKrypton:
> @AxelKrypton In which order should we merge the PRs? I can also include the suggestions by @RobinSattler and merge this PR first.

Of course, my PR #92 will conflict with this. Honestly I think it is better if @LucasConstantin reviews that first and we merge that first, so that you can also profit from that mechanism if you need to introduce a new base config file based on the sampler version.

AxelKrypton:
> > @AxelKrypton In which order should we merge the PRs? I can also include the suggestions by @RobinSattler and merge this PR first.
> 
> Of course, my PR #92 will conflict with this. Honestly I think it is better if @LucasConstantin reviews that first and we merge that first, so that you can also profit from that mechanism if you need to introduce a new base config file based on the sampler version.

@nilssass Sorry for the late merge of #92, but I oversaw this task in my todo list and totally forgot about that. You should now be able to rebase on `develop` and fix your changes using the new approach. The README file in the ***configs*** folder provides more information about how the default base config files are meant to be used and named.

nilssass:
This PR is outdated, so I will close it and open a new one with the current changes.

PR #94, opened 2025-12-09
Add Python formatter

### LINKED ISSUE #83: Add python formatting and remove bare `except` statements

RobinSattler:
Resolves #83.

I added `Python` formatting to the codebase based on `autopep8` which is set up to ignore rule `E26` to have unified copyright statements for `Bash` and `Python` files. The version for the testing framework is fixed to `autopep8==2.3.2`. @AxelKrypton, should we relax this to `autopep8~=2.3.2` or is it fine this way?

I'm not entirely happy that `Print_Info '\nRun ' --emph 'Hybrid-handler format' ' to correctly format the codebase.'` will be printed twice when both `Bash` and `Python` files are not correctly formatted (this is output of the unit test functions). But I do not see an easy solution for this right now.

I also added `*venv*` to the `.gitignore` file.

### DIFF
--- a/bash/formatter.bash
+++ b/bash/formatter.bash
@@ -1,6 +1,6 @@
 #===================================================
 #
-#    Copyright (c) 2023-2024
+#    Copyright (c) 2023-2025
 #      Hybrid-handler Team
 #
 #    GNU General Public License (GPLv3 or later)
@@ -19,14 +19,24 @@ function Format_Codebase()
     fi
 }
 
-function Run_Formatting_Unit_Test()
+function Run_Formatting_Unit_Tests()
 {
     Ensure_That_Given_Variables_Are_Set_And_Not_Empty HYBRID_top_level_path
     source "${HYBRID_top_level_path}/tests/unit_tests_formatting.bash" || exit ${HYBRID_fatal_builtin}
     # The following variable definition is just a patch to be able to reuse the test code from here
     HYBRIDT_repository_top_level_path="${HYBRID_top_level_path}"
-    if Unit_Test__codebase-formatting; then
-        Print_Info 'The codebase was correctly formatted and the formatting test passes.'
+    local bash_format_correct='FALSE'
+    local python_format_correct='FALSE'
+    if Unit_Test__codebase-formatting-Bash; then
+        bash_format_correct='TRUE'
+    else
+        printf '\n'
+    fi
+    if Unit_Test__codebase-formatting-Python; then
+        python_format_correct='TRUE'
+    fi
+    if [[ "${bash_format_correct}" = 'TRUE' ]] && [[ "${python_format_correct}" = 'TRUE' ]]; then
+        Print_Info 'The codebase was correctly formatted and the formatting tests pass.'
     fi
 }
 
--- a/bash/formatter.bash
+++ b/bash/formatter.bash
@@ -11,53 +11,61 @@ function Format_Codebase()
 {
     local bash_formatter_found='TRUE'
     local python_formatter_found='TRUE'
+    local bash_format_succeeded='FALSE'
+    local python_format_succeeded='FALSE'
+    Ensure_That_Given_Variables_Are_Set_And_Not_Empty HYBRID_top_level_path
     if ! hash shfmt &> /dev/null; then
         Print_Error \
-            'Command ' --emph 'shfmt' ' not available.' \
-            'Please install it (https://github.com/mvdan/sh#shfmt).\n'
+            'Command ' --emph 'shfmt' ' not available. Please install it (https://github.com/mvdan/sh#shfmt)' \
+            'to format the ' --emph 'Bash' ' files of the codebase.\n'
         bash_formatter_found='FALSE'
+    else
+        shfmt -w -ln bash -i 4 -bn -ci -sr -fn "${HYBRID_top_level_path}" &> /dev/null \
+            || Print_Internal_And_Exit 'Bash formatter ' --emph 'shfmt' ' unexpectedly failed.'
     fi
     if ! hash autopep8 &> /dev/null; then
         Print_Error \
-            'Command ' --emph 'autopep8' ' not available.' \
-            'Please install it (https://pypi.org/project/autopep8).\n'
+            'Command ' --emph 'autopep8' ' not available. Please install it (https://pypi.org/project/autopep8)' \
+            'to format the ' --emph 'Python' ' files of the codebase.\n'
         python_formatter_found='FALSE'
-    fi
-    if [[ "${bash_formatter_found}" = 'TRUE' ]] && [[ "${python_formatter_found}" = 'TRUE' ]]; then
-        Ensure_That_Given_Variables_Are_Set_And_Not_Empty HYBRID_top_level_path
-        shfmt -w -ln bash -i 4 -bn -ci -sr -fn "${HYBRID_top_level_path}"
-        local list_of_python_files=(
-            "${HYBRID_python_folder}/"*.py
-            "${HYBRID_top_level_path}/tests/"**/*.py
-        )
-        for file in ${list_of_python_files[@]}; do
-            # Ignoring rule E26, i.e. not enforcing a single whitespace after the # of inline comments
-            autopep8 --ignore "E26" -i "${file}"
-        done
     else
-        Print_Fatal_And_Exit 'Unable to format the codebase. ' \
-            'Please install the missing requirements and run the formatting again.'
+        # Ignoring rule E265, i.e. not enforcing a single whitespace after the # of block comments
+        autopep8 --ignore "E265" --max-line-length 119 -i "${HYBRID_top_level_path}/"{python,tests}/**/*.py \
+            &> /dev/null || Print_Internal_And_Exit 'Python formatter ' --emph 'autopep8' ' unexpectedly failed.'
+        python_format_succeeded='TRUE'
+    fi
+    if [[ ${bash_formatter_found} = 'FALSE' ]] && [[ ${python_formatter_found} = 'FALSE' ]]; then
+        Print_Fatal_And_Exit 'Unable to format the codebase. Please install the missing requirements ' \
+            'and run the formatting again.'
+    elif [[ ${bash_formatter_found} = 'TRUE' ]]; then
+        # NOTE: The Bash unit test is run since it does an additonal line length check on top of what the formatter
+        #       does. The Python unit test will not fail after formatting, so it is redundant to run it.
+        if __static__Run_Bash_Formatting_Unit_Test; then
+            bash_format_succeeded='TRUE'
+        else
+            Print_Info -l -- ''
+        fi
+    fi
+    if [[ ${bash_format_succeeded} = 'TRUE' ]] && [[ ${python_format_succeeded} = 'TRUE' ]]; then
+        Print_Info 'The codebase was correctly formatted and the ' --emph 'Bash' ' formatting test passes.'
+    elif [[ ${bash_format_succeeded} = 'FALSE' ]] && [[ ${python_format_succeeded} = 'TRUE' ]]; then
+        Print_Info 'The ' --emph 'Python' ' files of the codebase have been properly formatted.'
+    elif [[ ${bash_format_succeeded} = 'TRUE' ]] && [[ ${python_format_succeeded} = 'FALSE' ]]; then
+        Print_Info 'The ' --emph 'Bash' ' files of the codebase were correctly formatted and the ' --emph 'Bash' \
+            ' formatting test passes.'
+    else
+        Print_Internal_And_Exit 'The codebase was not correctly formatted.'
     fi
 }
 
-function Run_Formatting_Unit_Tests()
+function __static__Run_Bash_Formatting_Unit_Test()
 {
     Ensure_That_Given_Variables_Are_Set_And_Not_Empty HYBRID_top_level_path
     source "${HYBRID_top_level_path}/tests/unit_tests_formatting.bash" || exit ${HYBRID_fatal_builtin}
     # The following variable definition is just a patch to be able to reuse the test code from here
     HYBRIDT_repository_top_level_path="${HYBRID_top_level_path}"
-    local bash_format_correct='FALSE'
-    local python_format_correct='FALSE'
-    if Unit_Test__codebase-formatting-Bash; then
-        bash_format_correct='TRUE'
-    else
-        printf '\n'
-    fi
-    if Unit_Test__codebase-formatting-Python; then
-        python_format_correct='TRUE'
-    fi
-    if [[ "${bash_format_correct}" = 'TRUE' ]] && [[ "${python_format_correct}" = 'TRUE' ]]; then
-        Print_Info 'The codebase was correctly formatted and the formatting tests pass.'
+    if ! Unit_Test__codebase-formatting-Bash; then
+        return 1
     fi
 }
 
--- a/bash/formatter.bash
+++ b/bash/formatter.bash
@@ -9,13 +9,34 @@
 
 function Format_Codebase()
 {
+    local bash_formatter_found='TRUE'
+    local python_formatter_found='TRUE'
     if ! hash shfmt &> /dev/null; then
-        Print_Fatal_And_Exit \
-            'Command ' --emph 'shfmt' ' not available, unable to format codebase.' \
-            'Please, install it (https://github.com/mvdan/sh#shfmt) and run the formatting again.'
-    else
+        Print_Error \
+            'Command ' --emph 'shfmt' ' not available.' \
+            'Please install it (https://github.com/mvdan/sh#shfmt).\n'
+        bash_formatter_found='FALSE'
+    fi
+    if ! hash autopep8 &> /dev/null; then
+        Print_Error \
+            'Command ' --emph 'autopep8' ' not available.' \
+            'Please install it (https://pypi.org/project/autopep8).\n'
+        python_formatter_found='FALSE'
+    fi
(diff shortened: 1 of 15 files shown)


bash/formatter.bash
```diff
@@ -9,24 +9,55 @@
...
+            'Please install it (https://github.com/mvdan/sh#shfmt).\n'
+        bash_formatter_found='FALSE'
+    fi
+    if ! hash autopep8 &> /dev/null; then
+        Print_Error \
+            'Command ' --emph 'autopep8' ' not available.' \
+            'Please install it (https://pypi.org/project/autopep8).\n'
+        python_formatter_found='FALSE'
+    fi
+    if [[ "${bash_formatter_found}" = 'TRUE' ]] && [[ "${python_formatter_found}" = 'TRUE' ]]; then
```
AxelKrypton: Why did you put this in `&&`? If I have the Bash formatter but miss the Python one or vice-versa I do not get any formatting done. 🤔 Is this what we want? Suppose I have not installed `autopep8` and changed only Bash code. Then I cannot use `./Hybrid-handler format` because of that... I find this a bit rude, isn't it? 😅 

Also, the `HYBRID_python_folder` variable should be added to the `Ensure_That_Given_Variables_Are_Set_And_Not_Empty` call.

bash/formatter.bash
```diff
@@ -9,24 +9,55 @@
...
         Ensure_That_Given_Variables_Are_Set_And_Not_Empty HYBRID_top_level_path
         shfmt -w -ln bash -i 4 -bn -ci -sr -fn "${HYBRID_top_level_path}"
+        local list_of_python_files=(
+            "${HYBRID_python_folder}/"*.py
+            "${HYBRID_top_level_path}/tests/"**/*.py
```
AxelKrypton: The `**` in filename extension requires `globstar` option to be enabled. We should decide wether to enable it globally or locally. If you want to do it locally then a safe way to do so is:
```bash
    local -r string_to_restore_globstar=$(shopt -p globstar || true)
    shopt -s globstar
    local list_of_python_files=( ... )
    eval "${string_to_restore_globstar}" # NOTE: This eval usage is fine
```
I now saw that in the Bash unit test to check lines length (`shfmt` does not do so) does not set the option. Could you fix that, too, please?

bash/formatter.bash
```diff
@@ -9,24 +9,55 @@
...
+        )
+        for file in ${list_of_python_files[@]}; do
+            # Ignoring rule E26, i.e. not enforcing a single whitespace after the # of inline comments
+            autopep8 --ignore "E26" -i "${file}"
```
AxelKrypton: Fine to ignore `E26`, probably fine to impose it, too. In any case, does `autopep8` fail in some cases? because you let the handler exit in that case. 🤔 And in general, does the formatter print information or anything to output? I did not try your code yet.

I have a further question. Did you customise the accepted line length? I am pretty sure the default is 80, but in Bash we accept 120. So it is probably worth doing the same.

bash/formatter.bash
```diff
@@ -9,24 +9,55 @@
...
+        done
+    else
+        Print_Fatal_And_Exit 'Unable to format the codebase. ' \
+            'Please install the missing requirements and run the formatting again.'
     fi
 }
 
-function Run_Formatting_Unit_Test()
+function Run_Formatting_Unit_Tests()
```
AxelKrypton: Is it needed to run the Python unit test about formatting after having formatted the codebase? For Bash it is needed because the unit test checks things on top of what the formatter does (i.e. line length). If the Python formatting test cannot fail once the formatting has been done, then there is no reason to run it, isn't it? 🤔 Or do we foresee this possibly changing at some point?

You could rename the function to `Run_Bash_Formatting_Unit_Test` if we do not need to run Python test here.

tests/unit_tests_formatting.bash
```diff
@@ -51,19 +50,58 @@ function Unit_Test__codebase-formatting()
...
+    if hash autopep8 &> /dev/null; then
+        formatter_found='TRUE'
+    else
+        Print_Error 'Command ' --emph 'autopep8' \
+            ' not available, unable to fully check codebase formatting.'
+    fi
+    local list_of_python_files files_with_wrong_formatting file
+    list_of_python_files=(
+        "${HYBRIDT_repository_top_level_path}/python/"*.py
+        "${HYBRIDT_repository_top_level_path}/tests/"**/*.py
```
AxelKrypton: Maybe it is a good idea to add a function for preliminary operations and set there `globstar` and restore the previous state in the function automatically run after the unit test. If you do this for Bash then you can reuse those for Python. Ask if this is not clear.

tests/python_requirements.txt
```diff
@@ -1,5 +1,6 @@
 # Even if some requirements here are the same as for the handler, we repeat them
 # as in unit tests the handler code checking requirements does not necessarily run.
 
+autopep8==2.3.2
```
AxelKrypton: I reply here to your question. Actually `==` is much clearer than `~=` when there is a patch version number. Since `~=` means compatible with, and as it is explained [here](https://peps.python.org/pep-0440/#compatible-release) for patch numbers it means any following patch is accepted. I find this pretty strict and usually I use only minor numbers with `~=`. You can change the other requirements to `==` and keep `autopep8` so.

docs/developer/contributing.md
```diff
@@ -130,3 +130,7 @@ In particular, be aware the the formatter will not enforce the rules explained b
 
     * Indentation is done _exclusively with spaces_ and **no** <kbd>Tab</kbd> should be used.
     * Loops and conditional clauses are started on a single line, i.e. the `do` and `then` keywords are **NOT** put on a separate line.
+
+!!! info "Good to know about formatting Python files"
+    * `autopep8` is set up to intentionally ignore rule `E26`, i.e. not enforcing a single whitespace after the `#` of inline comments as recommended by the [PEP 8 – Style Guide for Python Code](https://peps.python.org/pep-0008/#inline-comments).
+      This is done to have unified copyright statements for Bash and Python files.
```
AxelKrypton, line 136: Maybe add: *"Everybody is encouraged to use a space after `#` in normal comments in Python files."*

AxelKrypton (changes requested):
A very nice starting point. I have some comments and possible improvements, plus I found a bug that actually I introduced myself about `**` in filename expansion. 🙈 If something is unclear, let's discuss. About what you mentioned in the description on having the message possibly printed twice, I am not sure we need to run the Python formatting unit test as part of the `format` execution mode and this would solve that. But in case we do need it, I will think more about it.

bash/formatter.bash
RobinSattler: I think ignoring `E26` makes sense to stay consistent with `shfmt` formatting style which does also not enforce a single whitespace after a `#`, if I'm not mistaken. `autopep8` does not print anything to the terminal and as far as I can see would only fail if no file is provided in this setting. Should we build in a catch for an `autopep8` failure? 🤔 

Good point, I will add the `--max-line-length 120` flag to the `autopep8` call. 👍

bash/formatter.bash
RobinSattler: True, I will adapt this. Should there be a `Print_Warning` or `Print_Attention` then for each missing formatter?

bash/formatter.bash
RobinSattler: Excellent point, no the unit test is not needed. I will rework this part. :+1:

bash/formatter.bash
RobinSattler: What are the pros and cons of enabling it globally vs. locally? :thinking:

bash/formatter.bash
AxelKrypton: Well, for each missing formatter you already give a `Print_Error` above, right? So I would say it is enough. 🤔

bash/formatter.bash
AxelKrypton: In general, developers have to be aware. For `globstar` actually we might even enable it globally, nothing comes to my head. I mean, it changes the meaning of `**` in filename expansion. Without `**` means `*` and I would say nobody should put in **on purpose** `**` without `globstar` enabled. So enabling it globally should not change anything. OK, in our case it does because we put in `**` thinking they were meaning the new thing, but this is our mistake... 🤪 

So maybe you could really enable it globally and mention this in the documentation, instead of doing it locally in the codebase and in the tests setup and teardown. 👍

bash/formatter.bash
AxelKrypton: > Should we build in a catch for an autopep8 failure?

Maybe a trailing `|| Print_Internal_And_Exit 'Python formatter autopep8 unexpectedly failed.'` 💪

bash/formatter.bash
AxelKrypton: Maybe you can then add an explanation comment in the execution mode where we call this function to say why we do not run the Python unit test (what I wrote here).

tests/python_requirements.txt
RobinSattler: Shouldn't we stay consistent in the testing framework with the requirements in _python/requirements.txt_? 🤔

bash/formatter.bash
```diff
@@ -9,24 +9,69 @@
 
 function Format_Codebase()
 {
+    local bash_formatter_found='TRUE'
+    local python_formatter_found='TRUE'
+    local bash_format_checked='FALSE'
+    local python_format_checked='FALSE'
```
AxelKrypton: I expected this flags to become true after having run the formatter, but this is not the case for Bash. Rename them to e.g. `{bash,python}_format_succeeded`.

bash/formatter.bash
```diff
@@ -9,24 +9,69 @@
...
-            'Command ' --emph 'shfmt' ' not available, unable to format codebase.' \
-            'Please, install it (https://github.com/mvdan/sh#shfmt) and run the formatting again.'
+        Print_Error \
+            'Command ' --emph 'shfmt' ' not available. Please install it (https://github.com/mvdan/sh#shfmt)' \
+            'to format the ' --emph 'Bash' ' files of the codebase.\n'
+        bash_formatter_found='FALSE'
     else
-        Ensure_That_Given_Variables_Are_Set_And_Not_Empty HYBRID_top_level_path
         shfmt -w -ln bash -i 4 -bn -ci -sr -fn "${HYBRID_top_level_path}"
+
```
AxelKrypton: I would fill this spare line with `|| Print_Internal_And_Exit 'Bash formatter ' --emph 'shfmt' ' unexpectedly failed.'` continuing the previous line by `\`. 🤪

bash/formatter.bash
```diff
@@ -9,24 +9,69 @@
...
+            "${HYBRID_top_level_path}/tests/"**/*.py
+        )
+        for file in ${list_of_python_files[@]}; do
+            # Ignoring rule E26, i.e. not enforcing a single whitespace after the # of inline comments
+            autopep8 --ignore "E26" --max-line-length 120 -i "${file}" \
+                || Print_Internal_And_Exit 'Python formatter ' --emph 'autopep8' ' unexpectedly failed.'
+        done
+        python_format_checked='TRUE'
+    fi
+    if [[ "${bash_formatter_found}" = 'FALSE' ]] && [[ "${python_formatter_found}" = 'FALSE' ]]; then
```
AxelKrypton: This is totally fine. Just as comment: Since they are strings

```bash
    if [[ "${bash_formatter_found}${python_formatter_found}" = 'FALSEFALSE' ]]; then
```

also works. Something to keep rather in mind for checking array of flags, though. Do not change it here. 😉

bash/formatter.bash
```diff
@@ -9,24 +9,69 @@
...
+        Print_Fatal_And_Exit 'Unable to format the codebase. Please install the missing requirements ' \
+            'and run the formatting again.'
+    elif [[ "${bash_formatter_found}" = 'TRUE' ]]; then
+        # NOTE: The Bash unit test is run since it does an additonal line length check on top of what the formatter
+        #       does. The Python unit test will not fail after formatting, so it is redundant to run it.
+        if __static__Run_Bash_Formatting_Unit_Test; then
+            bash_format_checked='TRUE'
+        else
+            printf '\n'
```
AxelKrypton: Strictly speaking this goes to standard output, while our logger does not. `Print_Info -l -- ''` should be almost equivalent (apart from invisible color codes). 😇

tests/unit_tests_formatting.bash
```diff
@@ -26,11 +26,10 @@ function Unit_Test__codebase-formatting()
     for file in "${list_of_source_files[@]}"; do
         if [[ $(wc -L < "${file}") -gt ${max_length} ]]; then
             files_with_too_long_lines+=("${file}")
-            continue
         fi
     done
     files_with_wrong_formatting=()
-    if [[ ${formatter_found} = 'TRUE' ]]; then
+    if [[ "${formatter_found}" = 'TRUE' ]]; then
```
AxelKrypton: This is kind of an own coding convention. I like not to quote booleans flags because if at some point for some reasons they get a space in them, code will break. I have no clue whether we are consistent with this, but I just wanted to say where it is coming from... It is impossible to enforce and probably this is also why I am pretty sure I did not mention it in the CONTRIBUTING file... 😅

tests/unit_tests_formatting.bash
AxelKrypton: I guess (consistently with the Bash test)

```bash
        "${HYBRIDT_repository_top_level_path}"/**/*.py
```

is enough unless you want to avoid match some files?!

tests/unit_tests_formatting.bash
```diff
@@ -51,19 +50,58 @@ function Unit_Test__codebase-formatting()
...
+    fi
+    local list_of_python_files files_with_wrong_formatting file
+    list_of_python_files=(
+        "${HYBRIDT_repository_top_level_path}/python/"*.py
+        "${HYBRIDT_repository_top_level_path}/tests/"**/*.py
+    )
+    files_with_wrong_formatting=()
+    if [[ "${formatter_found}" = 'TRUE' ]]; then
+        for file in "${list_of_python_files[@]}"; do
+            if $(autopep8 --ignore "E26" --max-line-length 120 -d --exit-code "${file}" &> /dev/null); then
```
AxelKrypton: Why do you use command expansion `$()` here? 🤔

AxelKrypton (approved):
Wonderful. I have one only major comment about a `$()` which tempted me to request changes, but I trust you and I will instead already approve, confident you will go through all comments. 😉 I left a couple of comments just FYI for future. You can simply read and resolve them.

tests/python_requirements.txt
AxelKrypton: It would be nice, but there is not really a (simple) way to enforce this. I mean, we could invent a way to let the tests runner pick in the tests python requirements some requirements from the top-level ***python*** folder, but I am not sure it is worth the hassle. At least not now. A nice enhancement, but well.... 😉

tests/unit_tests_formatting.bash
RobinSattler: Ah, thanks for catching this. True, it's not needed here. I will remove it.

tests/unit_tests_formatting.bash
RobinSattler: As discussed in private, this is to avoid matching some files, e.g., if someone uses an unhidden virtual environment folder, then this would be included by your proposition.

RobinSattler:
@AxelKrypton, thanks for all your helpful feedback! :muscle: One last thing, is there a convention how the copyright statements should look like? The _tests/tests_runner_ was only changed 2023. Should the copyright statement be altered to `Copyright (c) 2023-2025` or `Copyright (c) 2023, 2025`, i.e. do we exclude the year it was untouched?

Just to let you know, I will change the `--ignore "E26"` for `autopep8` to `--ignore "E265"` which is not that strict and only takes into account block comments.

tests/python_requirements.txt
RobinSattler: I didn't mean to automatize/unify this, but just keeping the requirements in both files the same. Otherwise, this might be "frustrating" for new users who turn into developers. Additionally, since the `autopep8` version will anyway only be checked if the unit test and not when `./Hybrid-handler format` is run, why not relax the requirement to at least include possible patches?

AxelKrypton:
> @AxelKrypton, thanks for all your helpful feedback! 💪 One last thing, is there a convention how the copyright statements should look like? The _tests/tests_runner_ was only changed 2023. Should the copyright statement be altered to `Copyright (c) 2023-2025` or `Copyright (c) 2023, 2025`, i.e. do we exclude the year it was untouched?

The correct statement is then `Copyright (c) 2023,2025` (without space). GPLv3 requests it so, with only the years where changes have been made, as far as I understood.

> Just to let you know, I will change the `--ignore "E26"` for `autopep8` to `--ignore "E265"` which is not that strict and only takes into account block comments.

Even better, thanks.

tests/python_requirements.txt
AxelKrypton: I meant to say that I am fine to keep the two sets of requirements consistent. Doing it by hand is never a guarantee, this was my point. If you want to keep `~= 2.3.2` allowing for following patches is totally fine, too. Looks a bit "unusual" to me, that's it.

PR #96, opened 2026-03-13
Add warning to doc if not the latest version and set `alias: true`

### LINKED ISSUE #63: Warn user who is not reading the latest documentation
According to [material-MkDocs](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=mike#version-warning) documentation it is possible to warn users in the documentation if they are on an old documentation version. I am in favour to add this to our documentation, too.

I would use this issue also to decide whether to set

```yaml
  version:
    alias: true
```

into [the MkDocs setup](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=mike#version-alias) to display the `latest` alias in the mike selector in the documentation.

RobinSattler:
This closes #63.

This PR implements the following:
- adds a warning if the user is not on the latest version (excluding `develop`)
- sets `alias: true`, i.e. the most recent release states `latest`
- fixes a dead link.

As already mentioned in the issue, some published documentations have to be re-deployed after the merge.

### DIFF
--- a/docs/developer/index.md
+++ b/docs/developer/index.md
@@ -7,7 +7,7 @@ hide:
 # From great power comes great responsibility
 
 The Hybrid-handler has been developed trying to make the entry point of a newbie developer as low as possible.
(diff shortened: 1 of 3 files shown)


AxelKrypton (approved):
The changes look as expected. As serving the doc locally does not render this feature and I'd need a sandbox as you did, I think it is fine to approve this and let you re-deploy the two documentations as we discussed (`2.0` and `2.1.3`). If you are not able to do so for permissions reasons or you want to do it together with 4 eyes looking at it, let's do it together. Then it will be obvious if something is to be fixed 😅 which I do not expect, though.

PR #97, opened 2026-03-16
Exchange forgotten old framework name with 'Hybrid-handler'

RobinSattler:
I noticed that I forgot to change the version title from `SMASH-vHLLE-hybrid` to `Hybrid-handler` in #85. The tags will be changed to `Hybrid-handler` from the next version on.

> [!IMPORTANT] 
> ~The version unit test will fail on local machines until the next release since `git describe` will return `SMASH-vHLLE-hybrid-2.1.3`.~ EDIT: Fixed by adapting the regex (see comments down below).

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,7 +1,7 @@
 cff-version: "1.2.0"
 message: "If you use this software, please cite it using these metadata."
 title: "smash-transport/smash-vhlle-hybrid: Hybrid-handler"
-version: "SMASH-vHLLE-hybrid-2.1.3"
+version: "Hybrid-handler-2.1.3"
 date-released: "2025-06-05"
 doi: "10.5281/zenodo.15880336"
 license: "GPL-3.0"
--- a/Hybrid-handler
+++ b/Hybrid-handler
@@ -2,7 +2,7 @@
 
 #===================================================
 #
-#    Copyright (c) 2023-2025
+#    Copyright (c) 2023-2026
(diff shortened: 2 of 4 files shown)


AxelKrypton (changes requested):
The changes are fine to me, but your admonition in the description leaves me a bit undecided, although it totally makes sense. I say undecided because I would in principle not really accept to merge this and make all developers have a failing test – in general. Now, here we are in a very small team and hopefully close to the next release. So this situation might be acceptable. 🤔 Alternatively for local tests green light one might
```diff
- HYBRID_codebase_version_regex='Hybrid-handler-[0-9]+([.][0-9]+)?'
+ HYBRID_codebase_version_regex='(Hybrid-handler|SMASH-vHLLE-hybrid)-[0-9]+([.][0-9]+)?'
```
and change this after the release. One could add a sentence to the comment, too. What's your take on this?

RobinSattler:
I see your point about the unit test with the `'(Hybrid-handler|SMASH-vHLLE-hybrid)-[0-9]+([.][0-9]+)?'` regex. A failing test is indeed not so nice, but on the other hand we might forget to remove this (which would not be so tragic after all). Since it's safer your way, I adapted the regex and added a comment.

AxelKrypton (approved):
I like your comment. Even if we forget to remove that for a while it will be harmless. If you really want to make it more probable not to forget, you can add a trivial Task-issue for this. 😅 Well done. 👍

PR #100, opened 2026-04-21
Avoid using bc command

### LINKED ISSUE #43: Add bc as requirement or do not use it
We do use `bc` command, but this is not listed as requirement. As we require e.g. `awk` already, maybe we can avoid using `bc`. If `bc` is not available some tests fail.

AxelKrypton:
This closes #43.

I first decided to add `bc` as requirement, but then since this might not be installed I changed mind and went for using `awk`, instead, which is already a requirement. The risk of using commands that are not requirements is always there and it is already emphasised in the developer documentation.

On the way I improved a couple of aspects, see commit messages.

@RobinSattler No hurry for the review, just by the milestone deadline whenever you find time.

### DIFF
--- a/tests/unit_tests_system_requirements.bash
+++ b/tests/unit_tests_system_requirements.bash
@@ -96,7 +96,7 @@ function Unit_Test__system-requirements()
     sed_version=4.2.1
     tput_version=5.9
     yq_version=4.24.2
-    Call_Codebase_Function_In_Subshell Check_System_Requirements
+    Call_Codebase_Function_In_Subshell Check_System_Requirements_And_Exit_If_Any_Is_Missing
     if [[ $? -ne 0 ]]; then
         Print_Error "Check system requirements of good system failed."
         return 1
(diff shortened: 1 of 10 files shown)


tests/unit_tests_system_requirements.bash
```diff
@@ -121,6 +133,7 @@ function Unit_Test__system-requirements()
     printf '\n'
     (
         unset -v 'TERM'
+        export PATH=${PATH/\/usr\/bin/} # Remove /usr/bin from PATH
```
RobinSattler, line 136: Is it enough to just remove `/usr/bin`? Wouldn't this test fail if I had installed all the requirements somewhere else (which might be also in `PATH`)? 🤔 Why not `unset` the `PATH`?

RobinSattler (approved):
I'm not really familiar with `awk`, but it seems to do the same thing as `bc` did the way you implemented it. I have one question, apart from that it looks good and I already approve! 🚀

tests/unit_tests_system_requirements.bash
AxelKrypton: This test only fails if something unexpected happens e.g. at Bash level, not if requirements are not found. The function is producing the report and that's its task. The idea of removing `/usr/bin` form `PATH` is to make more red crosses appear in the report and hence cover more code. I renamed the other function to make it clearer that the other exits and this one does not.

Unsetting all `PATH` is a bad idea because then Bash does not find anything any more. 🙈 Then the test might really fail. Already removing that folder forced me to set before `grep` path and "mock" it to keep it findable (it is usually in `/usr/bin`). That said, this test might be different on different OS, but I believe it is quite standard what we assume here.

RobinSattler: Makes sense, thanks for clarifying! :+1:

PR #101, opened 2026-04-22
Update configs with MAP parameters

### LINKED ISSUE #72: Update base configs with MAP parameters
@NGoetz recently made a Bayesian analysis for the hybrid [2503.10181]. I put the config file with them on our internal *lustre/AG-Elfner* directory. Currently there is some unfinished work in some modules, we should update the base configs when the they become publicly compatible.

RenHirayama:
Most changes are from adding another base config to keep the parameters used in Anna's paper, and making them consistent with the updated base configs.

Regarding the CHANGELOG message about the vHLLE branch, if it is merged before we tag, that part can be removed.

This closes #72.


configs/predef_configs/config_AuAu_4.3.yaml
```diff
@@ -22,6 +26,12 @@ Hydro:
...
     Executable: /path/to/hadron-sampler
+    Software_keys:
+        bulk: 0
+        ecrit: 0.5
 
 Afterburner:
     Executable: /path/to/smash
+    Software_keys:
+        Collision_Term:
+            Cross_Section_Scaling: 1
```
AxelKrypton: Missing EOL.

configs/vhlle_hydro
```diff
@@ -1,30 +1,41 @@
...
+etaSMin              0.2214
+etaSShiftMuB         0.1379
+etaSScaleMuB         2.8399
+
+zetaSparam           4
+zetaS                0.0816
+zetaSScaleBeta       0.103
+zetaSPeakEpsilon     24.039
+zetaSSigmaMinus      0.0725
+zetaSSigmaPlus       0.0378
```
AxelKrypton: Missing EOL.

docs/user/CHANGELOG/index.md
```diff
@@ -56,6 +56,10 @@ Given a version number `X.Y.Z`,
 
     :new: &nbsp; The `Input_file` configuration key in the `Hydro` and `Afterburner` modules can now be used to modify the expected file names from the appropriate :file_folder: ***IC*** and :file_folder: ***Sampler*** sub-folders, when a string without a `/` character is given.
 
+    :warning: &nbsp; The parameters extracted by the [Bayesian analysis](Phys. Rev. C 112, 014910) are now set as default for the base configs. The old parameters were kept in :file_folder: ***configs/predef_configs/vhlle_hydro_EPJA_58-11-230***.
```
AxelKrypton: Is this really the case? I mean, you added the *vhlle_hydro_EPJA_58-11-230*, but this contains the old parameters for the hydro part only, although you also adjusted the default configs for the other stages. 🤔

AxelKrypton (commented):
I left you some comments, but I believe that your changes should be cross-checked by someone else to ensure that the numbers are the correct ones. If @zpauliny cannot do this, please assign this work to another person in our group. This type of adjustments are error-prone, that's why I'd do this.

My main comment is about mentioning this change [in the user guide page about predefined configs](https://smash-transport.github.io/smash-vhlle-hybrid/develop/user/predefined_configs/), which is AFAIU not saying the truth any more, isn't it? 🤔

docs/user/CHANGELOG/index.md
RenHirayama: the main changes were in the hydro but I can improve the comment

RenHirayama:
> My main comment is about mentioning this change [in the user guide page about predefined configs](https://smash-transport.github.io/smash-vhlle-hybrid/develop/user/predefined_configs/), which is AFAIU not saying the truth any more, isn't it? 

I don't see what should be updated there, in particular the following is still true 
  
"They can be found in 📁 configs/predef_configs folder and the user is only required to insert the paths of the executables in the individual software sections. They can be executed in the standard manner using -c option in the execution mode of the handler."

docs/user/CHANGELOG/index.md
AxelKrypton: About the CHANGELOG, I would phrase it this way. The addition at the end clarifies my confusion in the review... I did not correctly get what you did.

```diff
The parameters extracted by the [Bayesian analysis](Phys. Rev. C 112, 014910) are now set as default
- for the base configs. The old parameters for the hydro evolution were kept in :file_folder:
- ***configs/predef_configs/vhlle_hydro_EPJA_58-11-230***, and the other parameters are
- adjusted by each handler config.
+ for the base configs. The old parameters for the hydro evolution were kept in :material-file:
+ *configs/predef_configs/vhlle_hydro_EPJA_58-11-230*, and the other parameters are
+ adjusted by each handler configuration file in :file_folder: *** predef_configs***.
```

AxelKrypton:
> > My main comment is about mentioning this change [in the user guide page about predefined configs](https://smash-transport.github.io/smash-vhlle-hybrid/develop/user/predefined_configs/), which is AFAIU not saying the truth any more, isn't it?
> 
> I don't see what should be updated there, in particular the following is still true

You're totally right pushing back on this. I hadn't understood what you did. 🤦‍♂️  I refined your phrasing in the CHANGELOG change unresolving a conversation.

Still I am wondering whether we should mention also elsewhere about the Bayesian parameters rather than only in the CHANGELOG?! 🤔

RenHirayama:
I don't know, we never mention the choice for base config keys

AxelKrypton:
> I don't know, we never mention the choice for base config keys

Yes, precisely. I noticed this as well. Maybe we should... 🤔

RenHirayama:
I put it in the `Config_File` key, which made sense to me

AxelKrypton:
> I put it in the `Config_File` key, which made sense to me

Wonderful, thanks for that. 🤩 I believe it is important to have it somewhere.

docs/user/CHANGELOG/index.md
AxelKrypton: I saw you added this point, thanks. Can you fix the first emoji and the italic without bold (as it is a file and not a folder) and the space after the triple `***` in the last folder (my bad, I had it here in my previous comment). 🙏

configs/vhlle_hydro
```diff
@@ -1,30 +1,41 @@
...
 
-tau0       2.7842
-tauMax     25.0
-dtau       0.05
+e_crit               0.3339
 
-freezeoutOnly 1
+etaSparam            3
+ah                   -10.9149
+al                   -10.5116
```
zpauliny: I believe these two: ah and al are switched

configs/vhlle_hydro
```diff
@@ -1,30 +1,41 @@
...
+T0                   0.1480
+etaSMin              0.2214
+etaSShiftMuB         0.1379
+etaSScaleMuB         2.8399
+
+zetaSparam           4
+zetaS                0.0816
+zetaSScaleBeta       0.103
```
zpauliny, line 38: I guess this is the conversion factor that I see explained in the paper in the paragraph about bulk viscosity? I cannot find that the numerical value is specified anywhere though, can you point me to it?

configs/predef_configs/vhlle_hydro_EPJA_58-11-230
```diff
@@ -0,0 +1,30 @@
...
+
+etaS       0.2
+etaSparam  0
+zetaS      0.0
+zetaSparam 0
+e_crit     0.5
+
+nx        161
+ny        161
+nz        71
```
zpauliny: If this is vhlle config that goes with previously studied energies in Anna's paper, can we maybe just use the large grid here?  nx: 201, ny: 201, nz: 401. And not recommend it in all individual configs. I understand that it is too big, but these configs are now not really the recommendation, more a collection of what was used previously.

zpauliny (approved):
Good work implementing this Renan, thanks! I have one general question: we would now recommend that people use the MAP parametrization, right? So maybe these predef configs for different energies with old parameters are a bit misleading, because it looks like we recommend: for AuAu @ 200 GeV, use this config. I would suggest grouping the old configs for different energies in a folder in predef configs called EPJA_58-11-230 or similar and document it properly. 

Other than that I found one mistake. I approve it now and maybe we can discuss whether there is a better way to store and point the user to the predef configs.

configs/vhlle_hydro
RenHirayama: these values were in a file that Niklas put together in `/lustre/hyihp/AG-Elfner/Utilities/hybrid-MAP-parameters.yaml`. Looking at the paper and the code, `zetaSScaleBeta` is indeed the beta in eq. 7 (duh) but I could not find the value in the paper

RenHirayama: nor in the thesis

PR #102, opened 2026-04-22
Constantin/change default vhlle input

LucasConstantin:
This PR updates the hybrid handler to support the renamed SMASH IC output (SMASH_IC_For_vHLLE.dat) while preserving compatibility with older versions (SMASH_IC.dat).
Hydro input is now selected based on the detected IC version.
If IC is not run, the handler gives a warning and applies the default for the newest version.

### DIFF
--- a/bash/Hydro_functionality.bash
+++ b/bash/Hydro_functionality.bash
@@ -25,7 +25,7 @@ function Ensure_All_Needed_Input_Exists_Hydro()
         "${HYBRID_software_input_file[Hydro]}"
     Ensure_Given_Files_Exist \
         "${HYBRID_software_configuration_file[Hydro]}"
-    Internally_Ensure_Given_Files_Exist "${HYBRID_software_output_directory[Hydro]}/SMASH_IC.dat"
+    Internally_Ensure_Given_Files_Exist "${HYBRID_software_output_directory[Hydro]}/SMASH_IC_For_vHLLE.dat"
 }
 
 function Ensure_Run_Reproducibility_Hydro()
@@ -41,7 +41,7 @@ function Run_Software_Hydro()
     cd "${HYBRID_software_output_directory[Hydro]}"
     local -r \
         hydro_config_file_path="${HYBRID_software_configuration_file[Hydro]}" \
-        ic_output_file_path="${HYBRID_software_output_directory[Hydro]}/SMASH_IC.dat"
+        ic_output_file_path="${HYBRID_software_output_directory[Hydro]}/SMASH_IC_For_vHLLE.dat"
     "${HYBRID_software_executable[Hydro]}" \
         "-params" "${hydro_config_file_path}" \
         "-ISinput" "${ic_output_file_path}" \
@@ -54,7 +54,7 @@ function Run_Software_Hydro()
 
 function __static__Create_Symbolic_Link_To_IC_File()
 {
-    local -r target_link_name="${HYBRID_software_output_directory[Hydro]}/SMASH_IC.dat"
+    local -r target_link_name="${HYBRID_software_output_directory[Hydro]}/SMASH_IC_For_vHLLE.dat"
     if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
         ln -s -f "${HYBRID_software_input_file[Hydro]}" "${target_link_name}"
     elif [[ ! "${target_link_name}" -ef "${HYBRID_software_input_file[Hydro]}" ]]; then
--- a/bash/Hydro_functionality.bash
+++ b/bash/Hydro_functionality.bash
@@ -25,7 +25,7 @@ function Ensure_All_Needed_Input_Exists_Hydro()
         "${HYBRID_software_input_file[Hydro]}"
     Ensure_Given_Files_Exist \
         "${HYBRID_software_configuration_file[Hydro]}"
-    Internally_Ensure_Given_Files_Exist "${HYBRID_software_output_directory[Hydro]}/SMASH_IC_For_vHLLE.dat"
+    Internally_Ensure_Given_Files_Exist "${HYBRID_software_output_directory[Hydro]}/SMASH_IC.dat"
 }
 
 function Ensure_Run_Reproducibility_Hydro()
@@ -41,7 +41,7 @@ function Run_Software_Hydro()
     cd "${HYBRID_software_output_directory[Hydro]}"
     local -r \
         hydro_config_file_path="${HYBRID_software_configuration_file[Hydro]}" \
-        ic_output_file_path="${HYBRID_software_output_directory[Hydro]}/SMASH_IC_For_vHLLE.dat"
+        ic_output_file_path="${HYBRID_software_output_directory[Hydro]}/SMASH_IC.dat"
     "${HYBRID_software_executable[Hydro]}" \
         "-params" "${hydro_config_file_path}" \
         "-ISinput" "${ic_output_file_path}" \
@@ -54,7 +54,7 @@ function Run_Software_Hydro()
 
 function __static__Create_Symbolic_Link_To_IC_File()
 {
-    local -r target_link_name="${HYBRID_software_output_directory[Hydro]}/SMASH_IC_For_vHLLE.dat"
+    local -r target_link_name="${HYBRID_software_output_directory[Hydro]}/SMASH_IC.dat"
     if [[ ! -f "${target_link_name}" || -L "${target_link_name}" ]]; then
         ln -s -f "${HYBRID_software_input_file[Hydro]}" "${target_link_name}"
     elif [[ ! "${target_link_name}" -ef "${HYBRID_software_input_file[Hydro]}" ]]; then
--- a/bash/global_variables.bash
+++ b/bash/global_variables.bash
@@ -36,15 +36,6 @@ function Define_Further_Global_Variables()
         [Add_corona_from_IC_and_Hydro]="${HYBRID_python_folder}/add_corona.py"
         [Latin_hypercube_sampling]="${HYBRID_python_folder}/latin_hypercube_sampling.py"
     )
-    declare -gA HYBRID_software_default_input_filename=(
-        [IC]=''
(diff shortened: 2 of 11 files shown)


bash/global_variables.bash
```diff
@@ -36,9 +36,9 @@ function Define_Further_Global_Variables()
         [Add_corona_from_IC_and_Hydro]="${HYBRID_python_folder}/add_corona.py"
         [Latin_hypercube_sampling]="${HYBRID_python_folder}/latin_hypercube_sampling.py"
     )
-    declare -rgA HYBRID_software_default_input_filename=(
+    declare -gA HYBRID_software_default_input_filename=(
         [IC]=''
-        [Hydro]="SMASH_IC.dat"
+        [Hydro]="" # This will be sat in the sanity checks as it depends on the IC executable version
```
AxelKrypton: * Use single quote as for `IC`.
* Typo: ~sat~ **set**.
* This array is not anymore constant, move it down to e.g. line 163.

bash/sanity_checks.bash
```diff
@@ -164,6 +165,42 @@ function __static__Set_Base_Configuration_File_If_Unset()
...
+    
+    case "${key}" in
+        Hydro)
+            # If IC is not being run, we can't determine its version to pick the right input filename
+            if ! Element_In_Array_Equals_To 'IC' "${HYBRID_given_software_sections[@]}"; then
+                Print_Warning \
+                    'It is not possible to deduce which input file should be used for the ' \
+                    --emph 'Hydro' ' stage, since the IC stage is not run. Falling back to default ' \
+                    --emph 'SMASH_IC_For_vHLLE.dat' '. Use the ' --emph 'Input_File' \
+                    ' key in the Hydro section to explicitly specify a filename.'
```
AxelKrypton, line 186: Nice. I would add some `\n` to make it easier to read in typical terminal size.

```suggestion
                Print_Warning \
                    'It is not possible to deduce which input file should be used for the ' \
                    --emph 'Hydro' ' stage,\nsince the IC stage is not run. Falling back to default ' \
                    --emph 'SMASH_IC_For_vHLLE.dat' '.\nUse the ' --emph 'Input_File' \
                    ' key in the Hydro section to explicitly specify a filename.'
```

tests/mocks/smash_IC_black-box.py
```diff
@@ -146,7 +146,13 @@ def finish():
     name_unfinished = ".unfinished"
     name_oscar = ".oscar"
     name_dat = ".dat"
+
     name_particles_file = "SMASH_IC"
```
AxelKrypton: Either keep this line or remove the `else:`.

tests/integration_tests_sanity_checks.bash
```diff
@@ -67,3 +68,39 @@ function Integration_Test__pick-correct-base-config()
         return 1
     fi
 }
+
+function Make_Test_Preliminary_Operations__pick-correct-default-input-file()
+{
+    Make_Test_Preliminary_Operations__pick-correct-base-config 'smash_IC_black-box.py'
+    HYBRID_given_software_sections=('IC' 'Hydro')
```
AxelKrypton, line 75: Correct, this is needed. Maybe add a comment to make it explicit:

```suggestion
    # Pretend sanity checks are done in a scenario where both IC and Hydro are run
    HYBRID_given_software_sections=('IC' 'Hydro')
```

tests/integration_tests_sanity_checks.bash
```diff
@@ -19,7 +20,7 @@ function Make_Test_Preliminary_Operations__pick-correct-base-config()
         source "${HYBRIDT_repository_top_level_path}/bash/${file_to_be_sourced}" || exit ${HYBRID_fatal_builtin}
     done
     Define_Further_Global_Variables
-    HYBRID_software_executable[IC]="${HYBRIDT_tests_folder}/mocks/echo.py"
+    HYBRID_software_executable[IC]="${HYBRIDT_tests_folder}/mocks/${ic_mock}"
```
AxelKrypton: I do not believe that this is needed. I do not remember why I told you that we had to call the other mock black-box script of IC. Here we do only care about a fake version split out from a mock executable and the simplest is `echo.py`. So I would get rid of the `local -r ic_mock=${1:-echo.py}` and simply restore the previous line here. When you call this function on line 74 do not pass in any argument. I tested it, it works.

tests/integration_tests_sanity_checks.bash
```diff
@@ -67,3 +68,39 @@ function Integration_Test__pick-correct-base-config()
         return 1
     fi
 }
+
+function Make_Test_Preliminary_Operations__pick-correct-default-input-file()
+{
+    Make_Test_Preliminary_Operations__pick-correct-base-config 'smash_IC_black-box.py'
```
AxelKrypton: Remove argument.

```suggestion
    Make_Test_Preliminary_Operations__pick-correct-base-config
```

tests/integration_tests_sanity_checks.bash
```diff
@@ -67,3 +68,39 @@ function Integration_Test__pick-correct-base-config()
...
+    HYBRID_given_software_sections=('IC' 'Hydro')
+}
+
+function __static__Test_Picked_Default_Input_File_For_Version()
+{
+    export MOCK_IC_VERSION="$1"
```
AxelKrypton: This should be `MOCK_ECHO_VERSION="$1"`.

tests/integration_tests_sanity_checks.bash
```diff
@@ -67,3 +68,39 @@ function Integration_Test__pick-correct-base-config()
...
+        [3.3]='SMASH_IC_For_vHLLE.dat'
+    )
+    local v
+    for v in "${!Hydro_cases[@]}"; do
+        (__static__Test_Picked_Default_Input_File_For_Version "${v}" "${Hydro_cases[${v}]}" 'Hydro')
+        if [[ $? -ne 0 ]]; then
+            return 1
+        fi
+    done
+}
```
AxelKrypton: Missind end of line.

bash/sanity_checks.bash
```diff
@@ -25,6 +25,7 @@ function Perform_Sanity_Checks_On_Provided_Input_And_Define_Auxiliary_Global_Var
             __static__Set_Software_Input_Data_File "${key}"
```
AxelKrypton: ⚠️ This must be moved down after your new function call!

AxelKrypton (changes requested):
@LucasConstantin Well done. 👍  I spotted few problems and you also need to fix the Hydro functionality unit test that is failing. You can use [this patch](https://github.com/user-attachments/files/27055511/Lucas.patch) downloading it and then using `git apply <patch file>` in your Hybrid handler folder. Or look its content and do it by hand, it is few changes only in one single file. Ask me if anything is unclear. 🙃

bash/global_variables.bash
LucasConstantin: Will do

tests/integration_tests_sanity_checks.bash
LucasConstantin: Indeed not needed

bash/Hydro_functionality.bash
AxelKrypton, line 1: I am not sure the changes in this file are correct. The `Hydro` output folder is containing a file whose name is internal to the Handler. That is the link to the `IC` output in the ***IC*** folder. That is the file that should be correctly named. By the way, in #105 I have extracted these internal names into global constants.

tests/mocks/smash_IC_black-box.py
```diff
@@ -146,7 +146,13 @@ def finish():
...
     name_oscar = ".oscar"
     name_dat = ".dat"
-    name_particles_file = "SMASH_IC"
+    name_spectators_file = "SMASH_IC"
+
+    from packaging.version import Version
+    if Version(ic_version) >= Version("3.3"):
+        name_particles_file = "SMASH_IC_For_vHLLE"
+    else:
+        name_particles_file = "SMASH_IC"
```
AxelKrypton, line 155: Are you sure about the two different names `name_spectators_file` and `name_particles_file`? I leave it up to you to make sure both variables are needed and used.

tests/unit_tests_Hydro_functionality.bash
AxelKrypton, line 1: Same comment here about the name of the file in the `Hydro` output directory. You should not change that. Instead, you should change those in the `IC` output directory, as you correctly did.

tests/unit_tests_scan_validation.bash
AxelKrypton, line 1: Why did you need these changes in this and the other scan_values test file? I saw also elsewhere that you quoted other keys, this is strictly speaking not needed, AFAIK. In the snipped below you see how Bash uses quotes in its answer when you ask to print the definition of an associative array in a way that can be reused for input:

```console
$ declare -A AAA=(['3.2']='Hello world' ['key with spaces']="Bye from $(whoami)")
$ declare -p AAA
declare -A AAA=(["key with spaces"]="Bye from asciarra" [3.2]="Hello world" )

```

AxelKrypton (changes requested):
@LucasConstantin well done, we are basically done. You still need to fix some file names that you changed but strictly speaking are not needed to be changed. I also left a couple of other minor comments to be sure there was no oversight.

tests/integration_tests_sanity_checks.bash
```diff
@@ -67,3 +67,40 @@ function Integration_Test__pick-correct-base-config()
...
+        Print_Error 'Picking default input file failed for ' --emph "${key}" ' and IC version ' --emph "$1" '.'
+        return 1
+    fi
+}
+
+function Integration_Test__pick-correct-default-input-file()
+{
+    declare -A Hydro_cases=(
+        ['3.2']='SMASH_IC.dat'
+        ['3.3']='SMASH_IC_For_vHLLE.dat'
```
AxelKrypton: For consistency, I'd remove single quotes on keys here, too.

AxelKrypton (approved):
Thanks @LucasConstantin for your work. I have a last minor comment, but I can surely already approve the changes. 🟢

PR #103, opened 2026-04-22
Add sanity check on number of IC events

### LINKED ISSUE #68: Implement sanity check for run with all hybrid stages and `Add_spectators_from_IC: true`

AxelKrypton:
This closes #68.

I left the existing check at "run time", but made it an internal failure.

### DIFF
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -171,10 +171,9 @@ function __static__Create_Sampled_Particles_File_Or_Symbolic_Link()
         else
             case $? in
                 ${HYBRID_fatal_value_error})
-                    exit_code=${HYBRID_fatal_value_error} Print_Fatal_And_Exit \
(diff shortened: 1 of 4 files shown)


LucasConstantin (approved):
This looks good to me

PR #105, opened 2026-04-24
Improve checks on stages functionality and extract internal names

### LINKED ISSUE #104: Check before creating symlinks and extract constant targets names
I discovered that `ln -s not-existing-file Ale` will succeed creating a broken link, i.e. `ln -s` does not check that the source exists. Therefore I suggest to check for source before creating symbolic links, e.g. in `Hydro` about `IC` and so on.

Furthermore, the name used for the target of such symbolic links is an internal implementation detail of the Hybrid-handler, although of course visible to the user. This is spread and hard-coded everywhere at the moment, both in code and in tests. We should collect those names in a constant global associative array. This will allow us to rename them anytime in a straightforward way.

AxelKrypton:
This closes #104.

@RenHirayama Although I have tried to be careful, this type of refactoring in Bash is difficult as stuff might break silently. Hopefully tests caught everything I did wrong... 😇 But it is also very late and I will be happy to go through your comments. 

I did not rename any file. Should we do it in this PR in your opinion?

### DIFF
--- a/bash/Afterburner_functionality.bash
+++ b/bash/Afterburner_functionality.bash
@@ -1,6 +1,6 @@
 #===================================================
 #
-#    Copyright (c) 2023-2025
+#    Copyright (c) 2023-2026
 #      Hybrid-handler Team
 #
 #    GNU General Public License (GPLv3 or later)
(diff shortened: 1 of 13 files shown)


RenHirayama (approved):
To test I had to adapt the `For_vHLLE` things by hand, but other than that the internal behavior works. Looking through the code I did not identify any problems.

PR #106, opened 2026-04-28
Fix missing space in `add_corona.py`

RenHirayama:
Bugfix: the ListModus looks for a string ` end ` (with a space at the end) to parse the end of an event.


RobinSattler (approved):
Looks good to me, although I did not test run this. As an alternative, we could also think about adapting SMASH to look for `" end"` instead of `" end "`.

I just jumped in because it was a short one. 😄

PR #108, opened 2026-05-07
Draft quick-start at beginning of user-guide

AxelKrypton:
@zpauliny @RenHirayama In this PR I tried to implement what we discussed this morning. From a perspective of a totally new user on a totally random OS, it is basically impossible to offer something solid. 😅 Take this as a proposal. I am very happy to implement your suggestion. If we can lower the newbie barrier and make their experience better, owe should definitely do it. You can `mkdocs serve` this branch and click on the blocks and on the annotations, but the top-level page appears like this. 😇 

image


docs/user/index.md
```diff
@@ -12,6 +12,54 @@ Once cloned the repository, you can simply run the `Hybrid-handler` script[^1].
...
+
+    ```bash
+    $ git clone https://github.com/smash-transport/smash-vhlle-hybrid.git
+    # some standard output (1)
+    $ ./smash-vhlle-hybrid/Hybrid-handler
+    # A nice, colorful report about system requirements (2)
+    ```
+
+    1. Git will clone using HTTPS protocol and locally create a :file_folder: ***smash-vhlle-hybrid*** folder.
+    2. :warning: At this point you will discover whether your OS is ready to go. If it is, you will see green version numbers and green ticks :white_check_mark: as in the image shown here below. If you see some red version number or red cross :x:, you need to take action. ![Image title](../images/Helper-example.png)
```
RenHirayama:
```suggestion
    2. :warning: At this point you will discover whether your OS is ready to go. If it is, you will see green version numbers and green ticks :white_check_mark: as in the image shown here below. If you see some red version number or red cross :x:, you need to install or update the corresponding requirement. ![Image title](../images/Helper-example.png)
```

docs/user/index.md
```diff
@@ -12,6 +12,54 @@ Once cloned the repository, you can simply run the `Hybrid-handler` script[^1].
...
+
+??? abstract "Once the OS is ready"
```
RenHirayama:
```suggestion
??? abstract "Setting up once the OS is ready"
```

docs/user/index.md
```diff
@@ -12,6 +12,54 @@ Once cloned the repository, you can simply run the `Hybrid-handler` script[^1].
...
+    $ cp smash-vhlle-hybrid/configs/predef_configs/config_TEST.yaml my-first-hybrid-run/hybrid.yaml
+    $ cd my-first-hybrid-run
+    $ SMASH_EXE='/path/to/smash'; HYDRO_EXE='/path/to/vHLLE'; SAMPLER_EXE='/path/to/hadron-sampler' # (2)!
+    $ sed -i -e "s|/path/to/smash|${SMASH_EXE}|g" -e "s|/path/to/vHLLE|${HYDRO_EXE}|g" -e "s|/path/to/hadron-sampler|${SAMPLER_EXE}|g" hybrid.yaml
+    # Optionally check that the replacement has happened correctly
+    $ grep 'Executable:' hybrid.yaml # (3)!
+    ```
+
+    1. :bulb: You will need to hand over the path to each executable to the Hybrid-handler through its configuration file, which we are setting up next.
+    2. :pencil: You need to replace these three paths to the actual three paths to your software:exclamation:
```
RenHirayama, line 47: I think we need only one of these

docs/user/index.md
```diff
@@ -12,6 +12,54 @@ Once cloned the repository, you can simply run the `Hybrid-handler` script[^1].
...
+    Alternatively to run these two highlighted lines, you can simply open the :material-file: _hybrid.yaml_ file and specify the paths there by hand at the 4 `Executable:` keys.
+    3. This should print the four executable paths in the `IC`, `Hydro`, `Sampler`, `Afterburner` order.
+
+??? tip "Finally running!"
```
RenHirayama:
```suggestion
??? tip "Running everything!"
```

RenHirayama (approved):
Looks very good!

I would rephrase a bit but it's up to you

docs/user/index.md
AxelKrypton: Since the first is more general and the second more technical, I think it does not harm to have both. I will keep them.

PR #109, opened 2026-05-08
Sass/spin

nilssass:
This PR adds support for the new spin and vorticity-related config keys.

For the sampler, the new keys are only enabled from version 3.3 on: `compute_spin_vector`, and `create_vorticity_vector_output`. I also added the new vorticity key to the Hydro config.

Additionally, I have also shifted the `hydro_coordinate_system` key to the version logic from v3.3 on, as @RobinSattler mentioned this in #93.

While doing this, I updated the sampler tests so they match the new version logic. I also ran into two unrelated test issues: shfmt complained about array keys like [3.2], so I quoted those, and the system requirements report failed on my machine because sort did not like the colored output. I fixed that by setting LC_ALL=C for that sort call.

### DIFF
--- a/bash/Sampler_functionality_SMASH.bash
+++ b/bash/Sampler_functionality_SMASH.bash
@@ -48,7 +48,8 @@ function Validate_Configuration_File_Of_SMASH()
     )
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
         allowed_keys+=('create_root_output')
-    elif Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.3'; then
+    fi
+    if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.3'; then
         allowed_keys+=('hydro_coordinate_system' 'compute_spin_vector' 'create_vorticity_vector_output')
     fi
 
--- a/bash/Sampler_functionality_SMASH.bash
+++ b/bash/Sampler_functionality_SMASH.bash
@@ -47,8 +47,11 @@ function Validate_Configuration_File_Of_SMASH()
         'ratio_pressure_energydensity'
     )
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
-        allowed_keys+=('create_root_output' 'hydro_coordinate_system')
+        allowed_keys+=('create_root_output')
+    elif Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.3'; then
+        allowed_keys+=('hydro_coordinate_system' 'compute_spin_vector' 'create_vorticity_vector_output')
     fi
+
     readonly allowed_keys
     local keys_to_be_found
     keys_to_be_found=4
@@ -88,7 +91,7 @@ function Validate_Configuration_File_Of_SMASH()
                 fi
                 ((keys_to_be_found--))
                 ;;
-            shear | bulk | create_root_output)
+            shear | bulk | create_root_output | compute_spin_vector | create_vorticity_vector_output)
                 if [[ ! "${value}" =~ ^[01]$ ]]; then
                     Print_Error 'Key ' --emph "${key}" ' must be either ' \
                         --emph '0' ' or ' --emph '1' '.'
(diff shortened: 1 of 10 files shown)


bash/Sampler_functionality_SMASH.bash
```diff
@@ -47,8 +47,12 @@ function Validate_Configuration_File_Of_SMASH()
         'ratio_pressure_energydensity'
     )
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.2'; then
-        allowed_keys+=('create_root_output' 'hydro_coordinate_system')
+        allowed_keys+=('create_root_output')
     fi
+    if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.3'; then
+        allowed_keys+=('hydro_coordinate_system' 'compute_spin_vector' 'create_vorticity_vector_output')
+    fi
```
AxelKrypton, line 54: This change makes me wonder about these keys in the config base file we ship. You do not want the user to use them with the default base file we ship? I mean, the file *configs/hadron_sampler__v3.2* is used from version `3.2` on, i.e. for `v3.3` as well. It does not contain these three keys and the user cannot then use these keys from the Hybrid-handler configuration file. 🤔 Is this intended or should a *configs/hadron_sampler__v3.3* file with the default of these three keys be added to this PR?

AxelKrypton (commented):
@nilssass I scrolled through the changes and have one comment only, but an important one. Can you check this, please?

bash/Sampler_functionality_SMASH.bash
nilssass: Thank you for pointing this out! I was indeed wondering how to proceed with the new keys and the base config. I am absolutely fine with adding a new base config for the hadron sampler. I was afraid it would get messy over time with new base configs. Would you recommend adding configs/hadron_sampler__v3.3?

AxelKrypton: Absolutely. It feels to me the right way to go. The syntax of the filename is fixed, see *[configs/README.md](https://github.com/smash-transport/smash-vhlle-hybrid/tree/develop/configs)* file.  @RobinSattler Do you agree?

RobinSattler: I agree, it's the most user-friendly approach to provide a new base config. @AxelKrypton, it's partly my fault that @nilssass went this way. We talked about it last week and I was apparently not in the right mind when I told him this was alright. :exploding_head: Thanks for catching it! :raised_hands:

AxelKrypton: No problem. @nilssass you know then what you need to do. 😉 By the way, this was only a comment, I let @Carl-Rosenkvist officially review this.

nilssass: Thank you, I have added the new sampler base config with all 3 new keys :)

Carl-Rosenkvist (approved):
Thanks for the update. I checked through the changes and the new sampler `3.3` suport together with the added validation/tests looks good to me.  Happy to approve.

docs/user/CHANGELOG/index.md
```diff
@@ -52,7 +52,7 @@ Given a version number `X.Y.Z`,
 
     :new: &nbsp; Added `Add_corona_from_IC_and_Hydro` configuration key in the `Afterburner` module, which merges the files containing corona particles from :file_folder: ***IC*** and :file_folder: ***Hydro*** into the particle lists sampled in :file_folder: ***Sampler***.
 
-    :new: &nbsp; Added new valid config key `hydro_coordinate_system` for SMASH sampler to handle hydrodynamic simulations in Cartesian coordinates properly, additional to the runs in tau-eta frame.
+    :new: &nbsp; Added new valid config key `hydro_coordinate_system`, `compute_spin_vector`, and `create_vorticity_vector_output` for SMASH sampler `3.3` to handle hydrodynamic simulations in Cartesian coordinates properly, additional to the runs in tau-eta frame.
```
RobinSattler: This seems not to be right. `compute_spin_vector` and `create_vorticity_vector_output` are not placed correctly in that sentence.

PR #110, opened 2026-05-12
Add Lucas to authors and update mailmap file

RobinSattler:
@LucasConstantin, please check if the info I added is correct (copied it from SMASH).

Also you should check out the first section _GitHub name and email address_ of this [Wiki page](https://gitlab.itp.uni-frankfurt.de/ag-elfner/group-wiki/-/wikis/SMASH/GitHub/GitHub-setup-and-workflow#github-name-and-email-address) since you committed with different email addresses.


LucasConstantin (approved):
Thanks Robin, everything is correct 👍

PR #111, opened 2026-05-19
Remove unused global variables

AxelKrypton:
Following up my message on Mattermost, I simply tried to remove the variables and nothing broke.

These changes will trigger in the future errors if attempting to use these variables and, therefore, it improves maintainability.


LucasConstantin (approved):
Looks good!

PR #113, opened 2026-05-19
Remove old framework name in version unit test

### LINKED ISSUE #98: Remove `SMASH-vHLLE-hybrid` from regex in version unit test during release of `2.2`
In PR #97 the regex to check the version in _tests/unit_tests_version.bash_ was altered to include `Hybrid-handler` as well as `SMASH-vHLLE-hybrid` (see code snippet down below and [line 23 to 25](https://github.com/smash-transport/smash-vhlle-hybrid/blob/679f38e34827612e41d1e0a0e743a46e17e93d39/tests/unit_tests_version.bash#L23-L25) of the bash file). Just changing it to `Hybrid-handler` would result in a failing version unit test when running tests on a local machine (or more precisely in a cloned repo) because `git describe` will return `SMASH-vHLLE-hybrid` until the next release. During (or after) the release `2.2`, `SMASH-vHLLE-hybrid` in the regex should be removed.

```bash
    # The regex includes 'SMASH-vHLLE-hybrid' to make this test 'backward-compatible'
    # until the next release version 2.2. Afterward, this should be removed.
    HYBRID_codebase_version_regex='(Hybrid-handler|SMASH-vHLLE-hybrid)-[0-9]+([.][0-9]+)?'
```

RobinSattler:
Closes #98

In PR #97 the regex to check the version in _tests/unit_tests_version.bash_ was altered to include `Hybrid-handler` as well as `SMASH-vHLLE-hybrid`. This was to maintain backward compatibility for the test until the `Hybrid-handler-2.2` release. This PR removes `SMASH-vHLLE-hybrid` from the regex.

### DIFF
--- a/tests/unit_tests_version.bash
+++ b/tests/unit_tests_version.bash
@@ -20,9 +20,7 @@ function Unit_Test__version()
     # be used as the repository is locally checked out effectively as a
     # black-box archive. Testing using a regex should cover all cases.
     HYBRID_codebase_version='Hybrid-handler-42.666.1'
-    # The regex includes 'SMASH-vHLLE-hybrid' to make this test 'backward-compatible'
-    # until the next release version 2.2. Afterward, this should be removed.
-    HYBRID_codebase_version_regex='(Hybrid-handler|SMASH-vHLLE-hybrid)-[0-9]+([.][0-9]+)?'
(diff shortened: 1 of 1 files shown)


AxelKrypton (approved):
Given tests pass, time to merge! 🟢

PR #114, opened 2026-05-19
Fix DOI in documentation to represent all versions

RobinSattler:
I just realized that I used the specific DOI for version `2.1.3` and not the one representing all versions in the documentation. This introduces the fix.


AxelKrypton (approved):
Thanks for fixing this. I am wondering whether it is possible and wether we should deploy the `2.2` documentation (and possibly the previous ones as well) with this commit. Otherwise people will probably cite the wrong DOI. 🤔

PR #115, opened 2026-05-27
Add recent features info and DynFlu prerequisites

zpauliny:
This is to update the documentation about the necessary software for DynFlu and should close #112. Take this as a draft @AxelKrypton, I count on our discussion here to improve it.

### DIFF
--- a/docs/user/prerequisites.md
+++ b/docs/user/prerequisites.md
@@ -74,7 +74,7 @@ To learn more about running it, consult [the relevant section](predefined_config
 | [SMASH](https://github.com/smash-transport/smash) | 3.3 or higher |
 | [vHLLE](https://github.com/yukarpenko/vhlle) | Tag `vhlle-smash-hybrid-2` |
 | [vHLLE parameters](https://github.com/yukarpenko/vhlle_params) | Tag `vhlle-smash-hybrid-1` |
-| [SMASH hadron sampler](https://github.com/smash-transport/smash-hadron-sampler) | 3.3 |
+| [SMASH hadron sampler](https://github.com/smash-transport/smash-hadron-sampler) | Same as SMASH[^2] |
 
 </div>
 ---
--- a/docs/user/prerequisites.md
+++ b/docs/user/prerequisites.md
@@ -34,6 +34,12 @@
     As SMASH is a dependency of the hadron sampler codebase, different versions of the latter have to be compiled with corresponding given versions of SMASH.
     However, if you need to use a different version of the hadron sampler software, this is probably not making a difference from the Hybrid-handler perspective and it is likely to work.
 
+!!! tip "Use more recent features"
+    If you plan to use recent features of the physics software, make sure you use the appropriate version of said software that might be different than the stated prerequisites.
+    Since Hybrid-handler is developed to provide the user with a complete chain in heavy ion collision modeling, an improvement in modeling can affect more than one of the software packages.
+    Check all the software for compatibility with the desired feature.
+    So far, we are highlighting **Dynamic Fluidization** as a recent feature and documenting its specific requirements here below.
+
 Instructions on how to compile or install the software above can be found at the provided links either in the official documentation or in the corresponding README files.
 
 !!! warning "Be consistent about dependencies"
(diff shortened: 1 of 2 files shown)


docs/user/prerequisites.md
```diff
@@ -56,6 +62,24 @@ Currently, this is realized for some of the different stages and the supported s
 </div>
 ---
```
AxelKrypton: I would remove this divider and keep it after the new section.

docs/user/prerequisites.md
```diff
@@ -34,6 +34,12 @@
     As SMASH is a dependency of the hadron sampler codebase, different versions of the latter have to be compiled with corresponding given versions of SMASH.
     However, if you need to use a different version of the hadron sampler software, this is probably not making a difference from the Hybrid-handler perspective and it is likely to work.
 
+!!! tip "Use more recent features"
+    If you plan to use recent features of the physics software, make sure you use the appropriate version of said software that might be different than the stated prerequisites.
+    Since Hybrid-handler is developed to provide the user with a complete chain in heavy ion collision modeling, an improvement in modeling can affect more than one of the software packages.
+    Check all the software for compatibility with the desired feature.
+    So far, we are highlighting **Dynamic Fluidization** as a recent feature and documenting its specific requirements here below.
```
AxelKrypton: I think that we should link to the new section. Naturally you can do it on _here below_ or you can rephrase dropping the second part of the sentence and linking on _Dynamic Fluidization_. As you like more. 😄

docs/user/prerequisites.md
```diff
@@ -56,6 +62,24 @@ Currently, this is realized for some of the different stages and the supported s
 </div>
 ---
 
+### Dynamic Fluidization
+
+The dynamic initial conditions setup described in [:newspaper: *Góes-Hirayama et al.: Phys.Rev.C 113 (2026) 4, 044906*](https://journals.aps.org/prc/abstract/10.1103/w87d-ps6f) requires more recent versions of the various software packages.
```
AxelKrypton: I think you have a trailing space on this line.

docs/user/prerequisites.md
```diff
@@ -56,6 +62,24 @@ Currently, this is realized for some of the different stages and the supported s
 </div>
 ---
 
+### Dynamic Fluidization
+
+The dynamic initial conditions setup described in [:newspaper: *Góes-Hirayama et al.: Phys.Rev.C 113 (2026) 4, 044906*](https://journals.aps.org/prc/abstract/10.1103/w87d-ps6f) requires more recent versions of the various software packages. 
+We note that it also uses a specific set of base configuration files.
+To learn more about running it, consult the relevant section of [The predefined configuration files](predefined_configs.md).
```
AxelKrypton: I think that I would rather link to the relevant section:

```suggestion
To learn more about running it, consult [the relevant section](predefined_configs.md#dynamic-fluidization-setup) of the predefined configuration files.
```

docs/user/prerequisites.md
```diff
@@ -56,6 +62,24 @@ Currently, this is realized for some of the different stages and the supported s
...
+
+<div class="center-table" markdown>
+
+| Physics software | Required version |
+| :----------------: | :--------------: |
+| [SMASH](https://github.com/smash-transport/smash) | 3.3 |
```
AxelKrypton: I guess this should be "3.3 or higher" as above.

docs/user/prerequisites.md
```diff
@@ -56,6 +62,24 @@ Currently, this is realized for some of the different stages and the supported s
...
+| [vHLLE](https://github.com/yukarpenko/vhlle) | Tag `vhlle-smash-hybrid-2` |
+| [vHLLE parameters](https://github.com/yukarpenko/vhlle_params) | Tag `vhlle-smash-hybrid-1` |
+| [SMASH hadron sampler](https://github.com/smash-transport/smash-hadron-sampler) | 3.3 |
```
AxelKrypton: For the hadron sampler we should decide what to do. Is FIST suitable for the dynamical fluidisation or the user should use SMASH? If so, I would specify SMASH explicitly in the last column, otherwise you could out there `[See above](#hadron-sampler)` to point to the section about the hadron sampler (if FIST is usable).

AxelKrypton (changes requested):
@zpauliny Good job. Overall I think the changes are complete. 💪  I would only add few cross links and make the prerequisites more precise in the table.

docs/user/prerequisites.md
RobinSattler: @zpauliny, correct me if I'm wrong, but according to Niklas and by the looks of the file _[include/HypersurfaceReader.h](https://github.com/vlvovch/fist-sampler/blob/cd93397750c85c7bfbd32a75628b7081cec24232/include/HypersurfaceReader.h)_ in the FIST sampler, it is not possible to use it for a hydrodynamical evolution that ran in Cartesian coordinates.

zpauliny: I am not sure if one can use `FIST`, I checked what @RobinSattler suggested here, that seems to suppose hyperbolic coordinates, so not good for DynFlu. I do not know the package that well, but I looked around and haven't found a parameter for switching to cartesian. Even if there was a possibility to use it, I did not test it, so I would recommend `smash-hadron-sampler` in the documentation. @RenHirayama, have you tried `FIST` with DynFlu?

zpauliny: So after checking in with @RenHirayama, FIST has not been tested and it is not clear if it can be used for DynFlu. Therefore I prefer sticking with `Same as SMASH` for the sampler in DynFlu requirements. I pushed this change in the table, @AxelKrypton if you are happy with this, you can click on the resolve and the PR should be ready for a merge :slightly_smiling_face:

AxelKrypton (approved):
Thanks @zpauliny for the cross-check and the update in the table. We can merge this and then I will deploy the develop documentation. 👍

PR #116, opened 2026-08-10
Add SMASH-3.3 IC config

RenHirayama:
I think this was forgotten in #87 / #105 but it's necessary. 

I dislike the increasing number of backwards compatible configs, and would rather create an `configs/old/` folder and just leave the most current version in `configs/` though. Might be relevant for #107


zpauliny (approved):
Great, thank you.

PR #117, opened 2026-08-10
Fix SMASH sampler spin config key and make it backward compatible

zpauliny:
When running the hybrid chain with the test config just updating the executables' paths, the sampler run fails and the log reports an unknown key `compute_spin_vector`. I think this is a wrong parameter name and it should be `sample_spin`. I have corrected it in the places I know of, please check if this is sufficient.

### DIFF
--- a/bash/Sampler_functionality_SMASH.bash
+++ b/bash/Sampler_functionality_SMASH.bash
@@ -50,7 +50,12 @@ function Validate_Configuration_File_Of_SMASH()
         allowed_keys+=('create_root_output')
     fi
     if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.3'; then
-        allowed_keys+=('hydro_coordinate_system' 'compute_spin_vector' 'create_vorticity_vector_output')
+        allowed_keys+=('hydro_coordinate_system' 'create_vorticity_vector_output')
+    fi
+    if Is_Version "${HYBRID_software_version[Sampler]}" -ge '3.4'; then
+        allowed_keys+=('compute_spin_vector')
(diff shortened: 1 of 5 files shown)


LucasConstantin:
@zpauliny thanks for finding this bug. We have decided to do a hybrid handler release that will contain the version dependent logic to be backwards compatible with SMASH-3.3 once we have tagged the new smash-hadron-sampler version 3.4.

nilssass:
However, this will not solve the problem for @zpauliny right now and it might take a while until we have the new release. So we should at least offer some solution even if it is just on `develop`. 
@zpauliny can you tell us how urgent you need the fix and if having it on develop would be working for you for now?

zpauliny:
The solution @LucasConstantin is proposing is fine with me. I now have a local fix and I agree that it is not super urgent to deal with 3.3 tag. It would be great however to run a complete hybrid with the test config with every release and we will easily catch these bugs.

RobinSattler:
@zpauliny, I think it makes sense to leave this PR open since all the interesting info is already in here and one of our group members just takes over. :slightly_smiling_face:

RobinSattler:
I just pushed a fix and @nilssass agreed to review it. Besides checking the code changes, it would be nice if you could do test runs with version `3.3` and the current `develop` of the sampler. Just to make sure that everything is working as expected.

nilssass (approved):
Thanks @RobinSattler for the fix. Everything is fine and implemented as we have discussed it above, so I approve the PR.

To add to my previous comment: I checked that all tests (unit, integration, functional) passed before approving the PR.

